Issue #42, opened 2013-01-23
createAccessToken() - Not expiring or returning existing tokens


F21:
I noticed that when we use the same `client_id` and `client_secret` to request a token on behalf of a user (say, f21@somecorp.com), a new token is generated even though a token already exists. In addition, the old token is not deleted.

I have been going through the spec, but could not find anything regarding this. 

Is this the correct behaviour? 

Cheers :)

PR #122, opened 2013-05-11
Ability to return the existing token

mmdumi:
Hello,
With my code i would like to fix this issue: https://github.com/bshaffer/oauth2-server-php/issues/42
I have added an option into server config to interrogate the db and return the existing token for the given client, user and scope.


csiszarattila:
Any chance to merge this feature in the future?

PR #203, opened 2013-07-31
Add redirect_status_code config param for AuthorizeController

### LINKED ISSUE #201: Configurable 3xx HTTP status code for redirect responses
http://tools.ietf.org/html/rfc6749#section-1.7

> While the examples in this specification show the use of the HTTP 302 status code, any other method available via the user-agent to accomplish this redirection is allowed and is considered to be an implementation detail.

This library, namely `OAuth2\Controller\AuthorizeController`, uses 302 for all redirects. I agree with 302 as the default since it matches the RFC 6749 examples, but I would prefer to use 303 in my auth server. A `redirect_status_code` configuration parameter would be nice.

If you agree, then I'd be happy to contribute the change.

ademarre:
Fixes #201


bdegoy:
It is likely that I do not understand the discussion, but why finally adopt 302? and why not 307?

PR #268, opened 2013-11-30
Implement jti for JWT Bearer tokens to prevent replay attacks.

### LINKED ISSUE #265: Implement 'jti' parameter for jwt to prevent replay attacks
The 'jti' parameter in a JWT is a nonce to prevent replay attacks. This is currently optional in the [spec](http://tools.ietf.org/html/draft-ietf-oauth-json-web-token-13#section-4.1.7) and is not currently implemented in the library.

We should add support for checking the jti to see if it has been used. If so, the request should be denied.

F21:
Closes #265.

In https://github.com/bshaffer/oauth2-server-php/issues/265#issuecomment-29544814 I mentioned that there is some uncertainty as to how "unique" we need our "jti" matching to be. 

With my implementation, that is now the responsibility of the library user. He decides how he wants to return the `jti` in the `getJti()` storage method.

### DIFF
--- a/src/OAuth2/GrantType/JwtBearer.php
+++ b/src/OAuth2/GrantType/JwtBearer.php
@@ -146,13 +146,14 @@ class JwtBearer implements GrantTypeInterface, ClientAssertionTypeInterface
         }
 
         // Check the jti (nonce)
+        // @see http://tools.ietf.org/html/draft-ietf-oauth-json-web-token-13#section-4.1.7
         if (isset($jwt['jti'])) {
-
             $jti = $this->storage->getJti($jwt['iss'], $jwt['sub'], $jwt['aud'], $jwt['exp'], $jwt['jti']);
 
             //Reject if jti is used and jwt is still valid (exp parameter has not expired).
             if ($jti && $jti['expires'] > time()) {
                 $response->setError(400, 'invalid_grant', "JSON Token Identifier (jti) has already been used");
+                
                 return null;
             } else {
                 $this->storage->setJti($jwt['iss'], $jwt['sub'], $jwt['aud'], $jwt['exp'], $jwt['jti']);
--- a/src/OAuth2/Storage/Mongo.php
+++ b/src/OAuth2/Storage/Mongo.php
@@ -292,10 +292,12 @@ class Mongo implements AuthorizationCodeInterface,
     public function getJti($client_id, $subject, $audience, $expiration, $jti)
     {
         //TODO: Needs mongodb implementation.
+        throw new \Exception('getJti() for the MongoDB driver is currently unimplemented.');
     }
 
     public function setJti($client_id, $subject, $audience, $expiration, $jti)
     {
         //TODO: Needs mongodb implementation.
+    	throw new \Exception('setJti() for the MongoDB driver is currently unimplemented.');
     }
 }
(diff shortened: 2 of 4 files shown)


TomHAnderson:
I may be a little late for the party, but while implementing this for Doctrine I see https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/GrantType/JwtBearer.php#L158 doesn't expire the Jti after it is used.  And if the Jti isn't found the Jwt is used without it.  

If a Jti is to be used only once then why does it's data stay the same after granting an auth token?
If a Jti is included at all why would you ignore it and fall back to Jwt?  
Is there a case where you only want to support Jwt calls with a Jti and exclude those without?

F21:
@TomHAnderson 

JTIs are optional as per the [spec](https://tools.ietf.org/html/draft-ietf-oauth-jwt-bearer-12#section-3). See point 7 in section 3.

The spec also mentions that the `jti` cannot be reused for until the `JWT` expires, so we need to keep it in the database.

I don't see where the `JTI` is being ignored.

The spec says that the `JTI` parameter is optional, so that's why all calls are allowed (whether they include a `JTI` or not). Maybe a configuration parameter can be implemented so that `JTI`s are required.

Hope that helps! :smile:

bshaffer:
@TomHAnderson,
@F21 is right on. We could definitely benefit from a config parameter `require_jti`, for instance, as yes I believe if you were to implement this measure you would want to enforce it across the board.

If you'd like to submit a PR, I'd start by converting the [`$audience`](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/GrantType/JwtBearer.php#L36) parameter to a `$config` array (preserving BC by checking if `$config` is a string) and adding a `require_jti` config parameter defaulting to `false`.

TomHAnderson:
@bshaffer That sounds ok though it doesn't cover expiring the Jti.  The audience could expand to 
```
Jit = { 
    "audience": {
        "require_jit": (bool) true,
        "audience": (string) url, etc,
        "expired": (int) unix datetime, 
    },
...
}
```
audience.audience is an indicator one or the other should be renamed.  I think because of this audience should not be turned into an array and instead a Jit.config should be created to handle customization of the Jit.

```
Jit = { 
    "audience": (string) value,
    "config": {
        "require_jit": (bool) true,
        "expired": (int) unix datetime, 
    },
...
}
```
Because we're building outside the spec (and could influence the spec to our design) I don't want to redefine what any existing value means.  

In the config ```require_jit``` is a boolean and if set to true will stop the validation of a Jwt should the Jit not validate.  

```expired``` would be a unix timestamp (pre the rest of the app) and would be set when the Jit is used.  If this value is set the Jit will not validate.

In order to enforce a Jit (where a Jwt without a Jit would validate) a server level configuration parameter of ```jwt``` would be passed to  ```OAuth2\Server->__construct``` of ```'jwt' => array ('require_jit' = true);```  Instead of a config of just requrie_jit putting it in the context of jwt makes a little more sense to me.

TomHAnderson:
This duplicates the jit, however.  Should require_jit be a config of Jwt?

bshaffer:
Why do you have an issue with the `JTI` expiring at the same time as the `JWT`? according to @F21 this is consistent with the spec.

I don't understand most of what you are proposing with the other parts of your comment, but in general I would want this:

```php
$config = array(
    'audience' => 'foo.com',
);

PR #351, opened 2014-03-18
Openid connect

bshaffer:
This is the motherload - I have built off the work done by @jsacksick and @bojanz and placed the code in its own directory, for organizational purposes.

I've added a lot more support in the `Server` class for these changes, and I've added a slew of additional tests.

Please let me know what you think.  It would be great to get this massive beast merged

### DIFF
--- a/src/OAuth2/Encryption/Jwt.php
+++ b/src/OAuth2/Encryption/Jwt.php
@@ -13,8 +13,8 @@ class Jwt implements EncryptionInterface
         $header = array('typ' => 'JWT', 'alg' => $algo);
 
         $segments = array(
-            $this->urlsafeB64Encode(json_encode($header)),
-            $this->urlsafeB64Encode(json_encode($payload))
+            $this->urlSafeB64Encode(json_encode($header)),
+            $this->urlSafeB64Encode(json_encode($payload))
         );
 
         $signing_input = implode('.', $segments);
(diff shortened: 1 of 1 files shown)


daedeloth:
Did anyone manage to get this working?
Some documentation or demo code would be greatly appreciated :)

belerophon:
Regarding the openssl_sign:
- I think you have to create a RSA key-pair for it to go away. the RSA key-pair is associated with the client application. Something like the following:

		$config = array(
            "digest_alg" => 'sha512',
            "private_key_bits" => 2048,
            "private_key_type" => OPENSSL_KEYTYPE_RSA,
        );

        // Create the private and public key
        $res = openssl_pkey_new($config);

        // Extract the private key from $res to $privKey
        $privKey = null;
        openssl_pkey_export($res, $privKey);

        // Extract the public key from $res to $pubKey
        $pubKey = openssl_pkey_get_details($res);
        $pubKey = $pubKey["key"];
        
        // insert the RSA key pair into the public key table
        // https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Storage/Pdo.php#L526
        $this->insert('{{%oauth_public_key}}', [
            'client_id' => 'YOU_CLIENT_APP_ID',
            'public_key' => $pubKey,
            'private_key' => $privKey,
            'encryption_algorithm' => 'RS512',
        ]);

It is possible to use other types of keys...

From my very, very limited knowledge, this is necessary when creating JWT tokens.
 What I still don't quite understand is why it is required when asking for openid scope, if no JWT token is given afterwards....

If someone can explain this better to me, I'd appreciate it also!! :)

Thanks!

daedeloth:
I think I've managed to get it working, but I'm really unsure if everything is implemented properly. I also have a very limited knowledge of what's happening, right now it looks like I've just implemented a basic oauth2 service with a user endpoint :/ But my openid client seems to be able to process the damn thing, so I guess it's fine.

I'll take a closer look at it after launch date :)

belerophon:
Can you share your overall configuration of the server, and the steps you've taken to make it work (e.g., did you create the key pair I've mentioned) ?

I am implementing this too, the server and clients, so I'd like to compare things...

daedeloth:
I have indeed created the private & public keys and stored those in the database table "oauth2_public_keys". client_id is your oauth2 client id.

My implementation is available here: https://github.com/CatLabInteractive/oauth2
Although I don't think it will be very useful since it's very much integrated in my framework.

Issue #338, opened 2014-02-21
"CryptoTokens" should match "JWT" spec


bshaffer:
CryptoToken is a word I made up. These are actually just JSON Web Tokens... why the distinction? Because I didn't know better.  Let's fix it!

bbalet:
Please update online documentation : 
http://bshaffer.github.io/oauth2-server-php-docs/overview/crypto-tokens/

bshaffer:
@bbalet I will as soon as the changes are merged into master for [v1.6](https://github.com/bshaffer/oauth2-server-php/pull/496)

bshaffer:
@bbalet docs updated and release tagged

PR #460, opened 2014-09-16
Rename JWT claims for JWT Access Tokens: #338

jaredk2g:
Hey all,

Thank you for creating such a fantastic library. It has been very useful to me. This pull request is my attempt to finish addressing #338 for JWT Access Tokens. A couple of notes here:
- renamed `client_id` to `aud`
- renamed `expires` to `exp`
- renamed `user_id` to `sub`
- added the `iss` parameter from an optional `issuer` configuration value
- added the `iat` parameter for the time the JWT access token was issued

I also added a method that maps the decoded access token from the JWT claims format to the format expected by other parts of this library. Hope this is helpful!
- Jared


Maks3w:
This changes need to be forwarded to https://github.com/bshaffer/oauth2-server-php-docs

PR #489, opened 2014-11-18
Set fetch mode to PDO::FETCH_BOTH in fetch and fetchAll calls

pedrosancao:
I'm current using this library in an application that already uses PDO for database connection, using object fetch mode. Both library and application uses the PDO connection and there is a fetch mode conflict.
To prevent these conflicts I updated the coding adding the fetch mode at `SELECT` calls.


aacotroneo:
Hi, I think PDO:: FETCH_ASSOC would be a better option than PDO::FETCH_BOTH as most of the interfaces that the PDO Storage implements state to return an associative array https://github.com/bshaffer/oauth2-server-php/pull/500. FETCH_BOTH brings unexpected additional data.

PR #491, opened 2014-12-03
http_build_query sometimes joins with "&amp;"

ghost:
I've no idea why, but http_build_query sometimes builds the query with &amp; as the joing string, rather than the proper default of &.
See http://php.net/manual/en/function.http-build-query.php#102324.

This is happening to me :(

Thanks!


Michal-sk:
This is an old issue, but ill mention it anyway. Make sure you check your ```php.ini``` for the ```arg_separator.output``` setting, that might be set to ```&amp;```

PR #499, opened 2015-01-16
Adding -H "Authorization:Basic " possibility

justonestep:
Just added the possibility to use the standard Basic Authorization header.

```PHP
base64_encode('ClientId,ClientSecret);
```

**Example Request**
```
curl -H "Authorization:Basic VGVzdENsaWVudCxUZXN0U2VjcmV0" https://api.mysite.com/token -d 'grant_type=client_credentials'
{"access_token":"03807cb390319329bdf6c777d4dfae9c0d3b3c35","expires_in":3600,"token_type":"Bearer","scope":null}
```

### DIFF
--- a/src/OAuth2/ClientAssertionType/HttpBasic.php
+++ b/src/OAuth2/ClientAssertionType/HttpBasic.php
@@ -101,6 +101,16 @@ class HttpBasic implements ClientAssertionTypeInterface
             return array('client_id' => $request->headers('PHP_AUTH_USER'), 'client_secret' => $request->headers('PHP_AUTH_PW'));
         }
 
+        if (!is_null($request->headers('Authorization'))) {
+          $auth = $request->headers('Authorization');
+          if (preg_match('/^Basic\s/', $auth)) {
+            $authBase = base64_decode(preg_replace('/^Basic\s/', '', $auth));
+            $data = explode(',', $authBase);
+
+            return array('client_id' => $data[0], 'client_secret' => $data[1]);
+          }
+        }
+
         if ($this->config['allow_credentials_in_request_body']) {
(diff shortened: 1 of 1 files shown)


Spomky:
The correct encoding is base64_encode(ClientId:ClientSecret);

And why do not use PHP_AUTH_USER, PHP_AUTH_PW, and AUTH_TYPE?

bshaffer:
This library already supports this:

https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/ClientAssertionType/HttpBasic.php#L100

Spomky:
This PR is a good idea (after the change above). 
This library should not rely only on the global variables. With some servers, they are not set and a check on the Authenticate header will fix this.

bshaffer:
@Spomky which web servers are you referring to which don't check this?

Spomky:
With PH-CGI or PH-FCGI for example.

bshaffer:
Okay, that seems reasonable.

@justonestep We would need a test case for this change, as well as changing the comment to a colon, as suggested by @Spomky. Also, if you could rebase it to be a single commit, that would be best.

Thank you for your contribution!

justonestep:
I changed the comment to a colon, added a test case and rebased it to an single commit
 (https://github.com/justonestep/oauth2-server-php/commit/baa30458abf258fe856ed9b4492ea619ef936431).
 Do I need to send a new pull request or is there another way to submit my changes?

bshaffer:
you could have kept the same PR open and just updated / rebased your branch. But you might as well submit a new PR now

justonestep:
I have no change to reopen this issue (button is deactivated "The feature/.. branch was force-pushed or recreated"). So I created a new [PR](https://github.com/bshaffer/oauth2-server-php/pull/501). Thanks!

PR #500, opened 2015-01-19
PDOStorage: pdo_fetch_mode changed to be FetchAssoc instead of FetchBoth

aacotroneo:
I think that PDO Storage should use FetchAssoc insetead of FetchBoth because:
  - 1 - Internally it only uses key values, never uses indexes directly.
  - 2 - Several of the interfaces implemented by the PDO Storage are expected to return an associative array (not a mixed array with indexes too). I was trying to print a token by doing a json_enconde but that does not work if fetch_mode is BOTH.

Regards!

### DIFF
--- a/src/OAuth2/Storage/Pdo.php
+++ b/src/OAuth2/Storage/Pdo.php
@@ -75,7 +75,7 @@ class Pdo implements
     {
         $stmt = $this->db->prepare(sprintf('SELECT * from %s where client_id = :client_id', $this->config['client_table']));
         $stmt->execute(compact('client_id'));
-        $result = $stmt->fetch(\PDO::FETCH_BOTH);
+        $result = $stmt->fetch(\PDO::FETCH_ASSOC);
 
         // make this extensible
         return $result && $result['client_secret'] == $client_secret;
@@ -86,7 +86,7 @@ class Pdo implements
         $stmt = $this->db->prepare(sprintf('SELECT * from %s where client_id = :client_id', $this->config['client_table']));
         $stmt->execute(compact('client_id'));
 
-        if (!$result = $stmt->fetch(\PDO::FETCH_BOTH)) {
+        if (!$result = $stmt->fetch(\PDO::FETCH_ASSOC)) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


pedrosancao:
PDO::FETCH_BOTH is the default behavior as stated in the [documentation](http://php.net/manual/pt_BR/pdostatement.fetch.php), but considering the memory saved by avoiding duplicated data seems a nice improvement.

aacotroneo:
Yep, the performance thing would be a plus, but I consider PDO's default (as sensitive as it may seem) an implementation detail. The point of the PR was that we are implementing an interface (e.g. AccessTokenInterface) that returns 'associative array', not a 'mixed array' as FETCH_BOTH does. Other implementations (Redis, CouchDb, Memory.. ) do return a pure associative array.

bshaffer:
I agree with this change... `FETCH_ASSOC` was implemented in #489 in order to explicitly state the fetch mode to be used. But as pointed out by @aacotroneo, there is no reason to fetch both.

Perhaps we should also include a config parameter for `fetch_mode`, default it to `FETCH_ASSOC`, and allow this to be changed (to prevent BC issues and also, apparently @pedrosancao has a reason for doing so, perhaps he just wanted to keep with the default)

aacotroneo:
FETCH_BOTH may be tricky to use, specially behind an interface. For example:
$a = some AccessTokenInterface() ..
echo count($a->getAccessToken());  
//prints 5 'almost always'
//prints 10 if it's a PDO Storage with FETCH_BOTH config

I think the problem @pedrosancao had was just that he reused a PDO connection which had FETCH_OBJ as a default. So, PDO Storage failed as it expects (at least) associative keys as results from the querys. Please, could you confirm this @pedrosancao?

It would be a BC for those expecting the access token (in my example) to have 10 columns, or numerical indexes. But numerical indexes are not supported by the interfaces anyway.

aacotroneo:
Anyway, I understand your concern about BC's @bshaffer so if you want I can add a fetch_type to the constructor configuration array like you suggested and make another PR. Right now the config only contains table names, I don't know if there's a better place.

Thanks!

pedrosancao:
Confirm, my connection used FETCH_OBJ as default, so the setting the default PDO behavior should fix this specific issue

bshaffer:
I'm okay with just making it `FETCH_ASSOC`... no reason to overcomplicate things.

PR #504, opened 2015-01-22
Master

jevylux:


bshaffer:
First of all you need to branch from develop and not master. Second of all I have no idea what this pull request is about.

PR #509, opened 2015-01-28
Fix #505 dynamodb value must not be empty

### LINKED ISSUE #505: "The value must not be empty" error message when using DynamoDB storage.
I am using the DynamoDB as storage. When I try to get an access token with grant_type 'client_credentials' or 'password',  I get the "The value must not be empty" error message. I traced  the error down to the 'setAccessToken' function of DynamodDB 

bshaffer\oauth2-server-php\src\OAuth2\Storage\DynamoDB.php

```
 public function setAccessToken($access_token, $client_id, $user_id, $expires, $scope = null)
    {
        // convert expires to datestring
        $expires = date('Y-m-d H:i:s', $expires);

        $clientData = compact('access_token', 'client_id', 'user_id', 'expires', 'scope');
        $clientData = array_filter($clientData, function ($value) { return !is_null($value); });

        $result = $this->client->putItem(array(
            'TableName' =>  $this->config['access_token_table'],
            'Item' => $this->client->formatAttributes($clientData)
        ));

        return true;

    }

```

It seems DynamoDB bombs out when i try to save an item with an attribute having an empty value.

It works if i replace "!is_null($value)" with "!empty($value) line 174 

```
$clientData = array_filter($clientData, function ($value) { return !empty($value); });
```

Same error when refreshing tokens,

```
public function setRefreshToken($refresh_token, $client_id, $user_id, $expires, $scope = null)
    {
        // convert expires to datestring
        $expires = date('Y-m-d H:i:s', $expires);

        $clientData = compact('refresh_token', 'client_id', 'user_id', 'expires', 'scope');
        $clientData = array_filter($clientData, function ($value) { return !is_null($value); });

        $result = $this->client->putItem(array(
            'TableName' =>  $this->config['refresh_token_table'],
            'Item' => $this->client->formatAttributes($clientData)
        ));

        return true;
    }
```

Is there a better solution/workaround this?  Thanks.

djhunx:
Changed filtering of client data by using empty() instead of is_null() in the setAccessToken, setAuthorizationCode and setRefreshToken methods of the DynamoDB storage..
Changed the getDefaultScope method to return null if there is no default scope set in the database.
Added tests for these changes.

### DIFF
--- a/src/OAuth2/Storage/DynamoDB.php
+++ b/src/OAuth2/Storage/DynamoDB.php
@@ -171,7 +171,7 @@ class DynamoDB implements
         $expires = date('Y-m-d H:i:s', $expires);
 
         $clientData = compact('access_token', 'client_id', 'user_id', 'expires', 'scope');
-        $clientData = array_filter($clientData, function ($value) { return !is_null($value); });
+        $clientData = array_filter($clientData, function ($value) { return !empty($value); });
 
         $result = $this->client->putItem(array(
             'TableName' =>  $this->config['access_token_table'],
@@ -208,7 +208,7 @@ class DynamoDB implements
         $expires = date('Y-m-d H:i:s', $expires);
 
         $clientData = compact('authorization_code', 'client_id', 'user_id', 'redirect_uri', 'expires', 'id_token', 'scope');
-        $clientData = array_filter($clientData, function ($value) { return !is_null($value); });
+        $clientData = array_filter($clientData, function ($value) { return !empty($value); });
 
         $result = $this->client->putItem(array(
             'TableName' =>  $this->config['code_table'],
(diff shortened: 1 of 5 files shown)


bshaffer:
two thoughts on this PR... and sorry it has taken me so long to review it...

1. If we DO want to pass a null value to dynamo DB (i.e. set the `client_secret` or `scope` value to null) will this implementation allow us to do this?

2. can you squash these changes into a single commit using `git rebase -i HEAD~3` ?

Thank you!

djhunx:
not a problem. thanks for looking into it.
 1.  we are using it now on a project we are working on. I'd say yes, this implementation allows us to let the client_secret and/or scope have null values. It fixes the situation when the scope is null and there's no default scope defined in the database.
 2.  i'm glad to squash the commits into one. (my first time to do it)

PR #524, opened 2015-02-20
Sign tokens with client_secret instead of private key for MAC-based algorithms

dtateii:
Adjusting for OpenID Connect id_token when using HS256, HS384, or HS512: http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation [3.1.3.7.8]


bshaffer:
I don't like this change.  If a client has a public key, it will be in the public key storage, and not in the client credential storage.

PR #534, opened 2015-03-06
Fixed some special chars in redirect uri while GrantType/AuthCode

tino415:
Authorized redirect uri was containing some unencoded chars...


bshaffer:
Hmm... shouldn't the decoding be done upon storage? I don't see why we would want this `url-encoded` in the database/storage...

tino415:
I agree, this was just quick fix, i just have problem that there was unencoded url, i probably should write bug report...

bshaffer:
Yes, let's try to fix this at the root. How did the *url-encoded* URL make it into the database in the first place?

bshaffer:
closing until a better solution is proposed

PR #552, opened 2015-03-27
fix got null sometime

WitzHsiao:
I found a bug when I get Access Token from DynamoDB, which will got null very some time.

### DIFF
--- a/src/OAuth2/Storage/DynamoDB.php
+++ b/src/OAuth2/Storage/DynamoDB.php
@@ -151,6 +151,7 @@ class DynamoDB implements
     public function getAccessToken($access_token)
     {
         $result = $this->client->getItem(array(
+            "ConsistentRead" => true,
             "TableName"=> $this->config['access_token_table'],
             "Key" => array('access_token'   => array('S' => $access_token))
         ));


bshaffer:
I feel like the `ConsistentRead` option could be supplied to the `DynamoDB` storage on creation. This would allow the user of the application to decide if they wanted to use consistent reads or not.

Eventual Consistency most likely would be acceptable in most cases.

WitzHsiao:
sounds like a plan!

bshaffer:
is this something you are willing to add in another Pull Request?

bshaffer:
@WitzHsiao are you planning on submitting a PR for this change?

WitzHsiao:
@bshaffer Sorry for reply lately. I've changed it to get ```ConsistenRead``` from config of ```DynamoDB```.

bshaffer:
These should all read

```php
"ConsistenRead" => $this->config['consistent_read'] 
```

With the `consistent_read` parameter being defined [here](https://github.com/WitzHsiao/oauth2-server-php/blob/develop/src/OAuth2/Storage/DynamoDB.php#L75) and defaulting to `false`

bshaffer:
Also, there's a typo. All of these say `ConsistenRead` instead of `ConsistentRead`.

Also, we need tests.

bshaffer:
@WitzHsiao pinging on this open issue

WitzHsiao:
@bshaffer Sorry, I forgot this post. Will fix it ASAP. Thanks for review.

bshaffer:
This is not what I had in mind. Rather than using `isset` each time, you can have the parameter defaulted in the constructor. i.e [here](https://github.com/WitzHsiao/oauth2-server-php/blob/develop/src/OAuth2/Storage/DynamoDB.php#L75):

```php
        $this->config = array_merge(array(
            'client_table' => 'oauth_clients',
            'access_token_table' => 'oauth_access_tokens',
            'refresh_token_table' => 'oauth_refresh_tokens',
            'code_table' => 'oauth_authorization_codes',
            'user_table' => 'oauth_users',
            'jwt_table'  => 'oauth_jwt',
            'scope_table'  => 'oauth_scopes',
            'public_key_table'  => 'oauth_public_keys',
            'consistent_read' => false, // <---- HERE
        ), $config);
```

PR #554, opened 2015-03-31
Makes possible to inherit from Oauth2\Request and use own Request class instead of original

hiqdev:
Makes possible to use own featured Request like this:

```php  
MyRequest extends \OAuth2\Request {
    public function someCoolFeature ($arg) {
        return $it;
    }
};

$request = MyRequest::createFromGlobals();

$server->handleTokenRequest($request);
```


bshaffer:
I love it! Thank you.

hiqdev:
Glad to hear it :)
Thank you for the library!

PR #559, opened 2015-04-10
fix: priority of client grant type validation in TokenController

Divi:
If the grant type is not available for the selected client, throw an exception before checking if the requested grant type (credentials, etc) is valid.
It can avoid requesting resources for nothing.


Divi:
Well, I will launch tests locally and will reopen this PR :)

bshaffer:
yes, there was a reason those functions happen in the order they did. I understand the order of some error messages seem a little unintuitive, but there wasn't a clean way to reorganize it, as far as I could tell.

But please give it a crack!! If the tests pass, I'll merge it :)

Issue #560, opened 2015-04-13
Resource request without access_token


davex69:
Hi,

Seems like I've found potential bug, when client make resource request with wrong `access_token` library validate it as https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Controller/ResourceController.php#L82

But when client make request without `access_token` `$responce->send()` return only 401 code of header and empty array. I think should validate this type of error.

bshaffer:
this is working as designed, the [TokenType](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/TokenType/Bearer.php#L86) is a `401` and no other information, as specified in the OAuth2 specification:

>  If the request lacks any authentication information (e.g., the client
   was unaware that authentication is necessary or attempted using an
   unsupported authentication method), the resource server SHOULD NOT
   include an error code or other error information.

PR #561, opened 2015-04-14
fix resource request without access_token

davex69:
When client make request without `access_token` $responce->send() return only 401 code of header and empty array. I think should validate this type of error.

Issue: https://github.com/bshaffer/oauth2-server-php/issues/560

### DIFF
--- a/src/OAuth2/Controller/ResourceController.php
+++ b/src/OAuth2/Controller/ResourceController.php
@@ -74,19 +74,19 @@ class ResourceController implements ResourceControllerInterface
     public function getAccessTokenData(RequestInterface $request, ResponseInterface $response)
     {
         // Get the token parameter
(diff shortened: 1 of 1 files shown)


bshaffer:
this is working as designed:

the [test](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/OAuth2/Controller/ResourceControllerTest.php#L13) is failing because the [TokenType](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/TokenType/Bearer.php#L86) is a `401` and no other information, as specified in the OAuth2 specification:

>  If the request lacks any authentication information (e.g., the client
   was unaware that authentication is necessary or attempted using an
   unsupported authentication method), the resource server SHOULD NOT
   include an error code or other error information.

PR #565, opened 2015-04-15
Add support for IbmDb2 (IBM i)

alanseiden:
Create storage class for the ibm_db2 extension, which is commonly used with DB2 on IBM i. The class includes DDL optimized for IBM i (chiefly the "for system name" clause).

### DIFF
--- a/src/OAuth2/Storage/IbmDb2.php
+++ b/src/OAuth2/Storage/IbmDb2.php
@@ -6,7 +6,6 @@ use OAuth2\OpenID\Storage\UserClaimsInterface;
 use OAuth2\OpenID\Storage\AuthorizationCodeInterface as OpenIDAuthorizationCodeInterface;
 
 /**
- * NOTE: WIP : convert pdo stuff to db2
  *
  * NOTE: This class is meant to get users started
  * quickly. If your application requires further
@@ -42,16 +41,9 @@ class IbmDb2 implements
                 throw new \InvalidArgumentException('First argument to OAuth2\Storage\IbmDb2 must be a resource or a configuration array');
             }
 
-            /* FYI: ZF2 used more flexible naming
-               $database = $findParameterValue(array('database', 'db'));
-               $username = $findParameterValue(array('username', 'uid', 'UID'));
-               $password = $findParameterValue(array('password', 'pwd', 'PWD'));
-              $isPersistent = $findParameterValue(array('persistent', 'PERSISTENT', 'Persistent'));
-           */
-
             // merge optional parameters. Set empty defaults if not present in $connection array.
             $connection = array_merge(array(
-                'db' =>     '',
+                'database' =>     '',
                 'username' =>   '',
                 'password' =>   '',
                 'persistent' => false,
@@ -127,11 +119,12 @@ class IbmDb2 implements
         // if it exists, update it.
         if ($this->getClientDetails($client_id)) {
             $stmt = db2_prepare($this->db, $sql = sprintf('UPDATE %s SET client_secret=?, redirect_uri=?, grant_types=?, scope=?, user_id=? where client_id=?', $this->config['client_table']));
+            return db2_execute($stmt, compact('client_secret', 'redirect_uri', 'grant_types', 'scope', 'user_id', 'client_id'));
         } else {
             $stmt = db2_prepare($this->db, sprintf('INSERT INTO %s (client_id, client_secret, redirect_uri, grant_types, scope, user_id) VALUES (?, ?, ?, ?, ?, ?)', $this->config['client_table']));
+            return db2_execute($stmt, compact('client_id', 'client_secret', 'redirect_uri', 'grant_types', 'scope', 'user_id'));
+            
         }
-
-        return db2_execute($stmt, compact('client_id', 'client_secret', 'redirect_uri', 'grant_types', 'scope', 'user_id'));
 
     }
 
@@ -157,9 +150,10 @@ class IbmDb2 implements
 
         if ($token = db2_fetch_assoc($stmt)) {
             
-            //replace 10th character (dash) between day and time with a space
+            // db2 timestamps look like yyyy-mm-dd-hh.mm.ss.000000 where the last six are microseconds.
+            // replace 10th character (dash between day and time) with a space to make it intelligible to strtotime()
             $token['expires'] = substr_replace($token['expires'], ' ', 10, 1);
-	    // convert date string back to timestamp
+	        // convert date string back to Unix timestamp
             $token['expires'] = strtotime($token['expires']);
         }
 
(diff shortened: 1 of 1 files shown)


src/OAuth2/Storage/IbmDb2.php
```diff
@@ -0,0 +1,606 @@
...
+            // use persistent or not
+            $isPersistent = $connection['persistent'];
+            $connectFunction = ((bool) $isPersistent) ? 'db2_pconnect' : 'db2_connect';
+
+            // try to connect
+            $connection = $connectFunction($connection['database'], $connection['username'], $connection['password'], $connection['driver_options']);
+
+            // this is how ZF2 handles connection errors
+            if ($connection === false) {
+                throw new Exception\RuntimeException(sprintf(
```
bshaffer: This will throw a namespace error, as it will be looking for `OAuth2\Storage\ Exception\RuntimeException`

bshaffer:
Wow, this is impressive! I am unfamiliar with IBM DB2... how can we add tests for this?

It is easy to add tests for new storage engines in the test suite... you just need to add the class to the [BaseTest::provideStorage](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/BaseTest.php#L7) method, and all the applicable storage tests will be run.

alanseiden:
It seems that Pdo's tests use sqlite; DynamoDb uses mock objects. Are there any tests with something similar to ibm_db2 that requires a real server? The ibm_db2 extension is not OO, so might not be easy to mock.

bshaffer:
@alanseiden DynamoDB doesn't use mock objects in the storage tests - it calls a real DynamoDB instance that I've set up with my Amazon credentials. I had to add  [an exception](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/Bootstrap.php#L510) so it only runs against php 5.5 in travis ci - otherwise the tests ran 6 times and conflicted with each other.

Is it possible to test against a local (travis) or remote (cloud) instance of IBM DB2?

src/OAuth2/Storage/IbmDb2.php
```diff
@@ -0,0 +1,606 @@
...
+                'database' =>     '',
+                'username' =>   '',
+                'password' =>   '',
+                'persistent' => false,
+                'driver_options' =>    array(),
+            ), $connection);
+
+            // use persistent or not
+            $isPersistent = $connection['persistent'];
+            $connectFunction = ((bool) $isPersistent) ? 'db2_pconnect' : 'db2_connect';
```
bshaffer: we should verify here that these functions exist - if they don't, we should throw an exception requiring the `ibm_db2` extension

src/OAuth2/Storage/IbmDb2.php
alanseiden: Would \Exception\RuntimeException (add slash in front) be acceptable here?

bshaffer:
Is there anything I can do to help this along? I am waiting to hear back from you concerning the testability of IMB-DB2

alanseiden:
IBM Db2 requires a real server. Clark Everetts of Zend wrote tests against a real IBM i server for ZF2's Zend_Db, so it's possible to do.

alanseiden:
I'm going to see Clark this week at a conference. Perhaps we can put our heads together.

bshaffer:
@alanseiden any progress on this?

alanseiden:
Today I did find a couple of bugs in the DDL and know how to correct the 
exception bug. Will correct those first.

Brent Shaffer wrote:

src/OAuth2/Storage/IbmDb2.php
bshaffer: no, the PHP class is `\RuntimeException` (see [here](http://php.net/manual/en/class.runtimeexception.php))

alanseiden: Yes, I tested \RuntimeException successfully today.

src/OAuth2/Storage/IbmDb2.php
alanseiden: I've added a check for the ibm_db2 extension using extension_loaded(), the technique used in ZF2. Sufficient or should we also check for the functions?

src/OAuth2/Storage/IbmDb2.php
```diff
@@ -0,0 +1,610 @@
...
+
+    /* OAuth2\Storage\ClientCredentialsInterface */
+    public function checkClientCredentials($client_id, $client_secret = null)
+    {
+        $stmt = db2_prepare($this->db, sprintf('SELECT * from %s where client_id = ?', $this->config['client_table']));
+        $successfulExecute = db2_execute($stmt, compact('client_id'));
+        $result = db2_fetch_assoc($stmt);
+
+        // make this extensible
+        return $result && $result['client_secret'] == $client_secret;
```
clarkphp: Is the purpose of the expression to test if $result exists, and if so, perform the == comparison?  The == comparison operator has higher precedence than && logical operator, meaning if I have the intent correct, the expression should actually be $result && ($result['client_secret'] == $client_secret).  In this way, PHP's short-circuit boolean evaluation can be applied without undefined variable notice. Or have I missed something obvious (it wouldn't be the first time)?

afilippov1985:
PDO is able to work with IBM DB2 http://php.net/manual/en/ref.pdo-ibm.php

bshaffer:
excellent find! We should add IBM DB2 to the travis tests then, and we can close this sucker.

alanseiden:
I don't understand how PDO would help with testing of the ibm_db2 extension.

bshaffer:
Using the [PDO storage class](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Storage/Pdo.php) and the `pdo-ibm` extension, a user should be able to integrate with this server and the IBM DB out of the box. They just need to create a `PDO` object using the [IMB_DB2 dsn](http://php.net/manual/en/ref.pdo-ibm.connection.php)

CiTRO33:
Hello @alanseiden
we are trying to use Oauth with JWT Access Token (https://bshaffer.github.io/oauth2-server-php-docs/overview/jwt-access-tokens/) . 

We insert the public / Private keys into the oauth_public_keys table.

If we want to generate a new access token,  we got an error.  
I figured out, that the SQL in the functions getPublicKey (Line 467 in /src/OAuth2/Storage/IbmDb2.php ), getPrivateKey  (Line 477 in /src/OAuth2/Storage/IbmDb2.php ) and getEncryptionAlgorithm  (Line 487 in /src/OAuth2/Storage/IbmDb2.php )  didn't work, because `ORDER BY client_id IS NOT NULL DESC`  didn't exists in DB2 V7R1.

When i change this to `ORDER BY client_id ASC` the correct data is returned. 
Sorted by example "testclient", NULL .

maybe its a performance improvements if we add `FETCH FIRST ROW ONLY` to retrun only one row.

Issue #564, opened 2015-04-15
Vulnerability in the JWT token validation


phindmarsh:
Recently a '[critical vulnerability](https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/)' with JWT was disclosed by Auth0.

It essentially boils down to modifying the algorithm header value and tricking a server using rsa signing to use the public key has an hmac secret instead, allowing an attacker who knows the servers public key to craft arbitrary tokens and have auth bypass. See the summary on the [Auth0 blog](https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/).

The solution to this is to pass the allowed algorithms when decoding a user-specified token, rather than trusting the value in the token header. This is usually fine since the server validating tokens should know what methods have been used to generate them.

The php-jwt lib has applied a patch for this which likely needs to be ported into the embedded JWT class of this library (see commits https://github.com/firebase/php-jwt/commit/10918f2a0181a53a2604c90b5904f9b33b0d3ba0 and https://github.com/firebase/php-jwt/commit/b2c2be6a45fda769c8c2ffe5ec4259a9d1e46e5b). Alternatively, this lib could depend on [firebase/php-jwt](https://github.com/firebase/php-jwt) directly. 

What are your thoughts on either updating the embedded lib, or depending on the php-jwt lib instead? I'm happy to submit a PR either way.

bshaffer:
Hey @phindmarsh! I do not take this lightly, and am sorry for not responding earlier. This is a critical change, and if you are truly willing to submit a PR, I would love to go with whichever option you find more appropriate. 

Patching the library itself would be easier, as bringing in `firebase/php-jwt` will require updating the documentation (because we would want it to be an optional dependency) and making other core updates. But, as you are more familiar with the problem itself, what is your preference?

phindmarsh:
No worries, I'm happy to push a PR for this today.

I agree about patching the library rather than pulling in `firebase/php-jwt`, although I do wonder if that is something to be considered for a later time (if only because that lib gets more eyes from a security context). Our OAuth server implementation uses it instead of the built in one at the recommendation of a security audit.

I'll get a PR together applying a fix for the internal JWT class, and we can move the migration to `firebase/php-jwt` to another discussion.

bshaffer:
Great! Yes, I agree it is ideal to use firebase's jwt library at some
point.

PR #567, opened 2015-04-21
Updated JWT lib to address #564

phindmarsh:
`JWT::decode()` now takes an array of `$allowed_algorithms` to prevent tricking the server into decoding using a different algorithm than intended.

The only required change here is specifying the `$allowed_algorithms` parameter to the `JWT::decode()` method and updating calls to that function elsewhere.

However since the JWT Grant Type doesn't store the RSA algorithm alongside the public key (and it looked like a bit of work to do that), I've extended the JWT class to provide supported RSA methods using `JWT::getSupportedMethods($type)` and refactored the `JWT::sign()` and `JWT::verifySignature()` methods to use the `JWT::$supportedMethods` declaration. 

This sounds like a big change, but its essentially a copy of the `firebase/php-jwt` implementation and it just seemed cleaner to do it this way. If you would rather we sticked to what was there and only change the `JWT::decode()` method signature, we'll just need to hard-code the allowed RSA methods in `JWTBearer::validateRequest()`. I figured doing that would be non-obvious for future maintenance but happy to follow your lead on whatever you feel comfortable with.

### DIFF
--- a/src/OAuth2/Encryption/EncryptionInterface.php
+++ b/src/OAuth2/Encryption/EncryptionInterface.php
@@ -4,8 +4,9 @@ namespace OAuth2\Encryption;
 
 interface EncryptionInterface
 {
+    public function getSupportedAlgorithms($type = null);
     public function encode($payload, $key, $algorithm = null);
-    public function decode($payload, $key, $algorithm = null);
+    public function decode($payload, $key, $allowed_algorithms = null);
     public function urlSafeB64Encode($data);
     public function urlSafeB64Decode($b64);
 }
--- a/src/OAuth2/Encryption/Jwt.php
+++ b/src/OAuth2/Encryption/Jwt.php
@@ -8,6 +8,16 @@ namespace OAuth2\Encryption;
  */
 class Jwt implements EncryptionInterface
 {
+
+    public static $supportedAlgorithms = array(
+        'HS256' => array('hash_hmac', 'sha256'),
+        'HS384' => array('hash_hmac', 'sha384'),
+        'HS512' => array('hash_hmac', 'sha512'),
+        'RS256' => array('openssl', 'sha256', 'OPENSSL_ALGO_SHA256'),
+        'RS384' => array('openssl', 'sha384', 'OPENSSL_ALGO_SHA384'),
+        'RS512' => array('openssl', 'sha512', 'OPENSSL_ALGO_SHA512')
+    );
+
     public function encode($payload, $key, $algo = 'HS256')
     {
         $header = $this->generateJwtHeader($payload, $algo);
@@ -25,7 +35,7 @@ class Jwt implements EncryptionInterface
         return implode('.', $segments);
     }
 
-    public function decode($jwt, $key = null, $verify = true)
+    public function decode($jwt, $key = null, $allowed_algorithms = null)
     {
         if (!strpos($jwt, '.')) {
             return false;
@@ -49,8 +59,12 @@ class Jwt implements EncryptionInterface
 
         $sig = $this->urlSafeB64Decode($cryptob64);
 
-        if ($verify) {
+        if (isset($key)) {
             if (!isset($header['alg'])) {
+                return false;
+            }
+
+            if (!in_array($header['alg'], (array) $allowed_algorithms)){
                 return false;
             }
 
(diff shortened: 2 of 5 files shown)


src/OAuth2/Encryption/Jwt.php
```diff
@@ -90,24 +100,17 @@ private function verifySignature($signature, $input, $key, $algo = 'HS256')
...
+                return hash_hmac($algorithm, $input, $key, true);
 
-            case 'RS384':
-                return $this->generateRSASignature($input, $key, defined('OPENSSL_ALGO_SHA384') ? OPENSSL_ALGO_SHA384 : 'sha384');
-
-            case 'RS512':
-                return $this->generateRSASignature($input, $key, defined('OPENSSL_ALGO_SHA512') ? OPENSSL_ALGO_SHA512 : 'sha512');
+            case 'openssl':
+                if(defined(self::$supportedAlgorithms[$algo][2])){
+                    $algorithm = constant(self::$supportedAlgorithms[$algo][2]);
```
bshaffer: can you explain exactly what's happening here? We're using the first array item as the algorithm, and the proceeding array items as arguments to the algorithm's function?

src/OAuth2/Encryption/Jwt.php
```diff
@@ -142,6 +145,25 @@ public function urlSafeB64Decode($b64)
...
 
+    public function getSupportedAlgorithms($type = null)
+    {
+        if ($type === null) {
+            return array_keys(self::$supportedAlgorithms);
+        }
+        else {
+            $filtered = array();
+            foreach (self::$supportedAlgorithms as $alg => $method) {
+                if ($type === 'RSA' && $alg[0] === 'R') {
```
bshaffer: creative way to filter these, but I can't find anything wrong with it.

src/OAuth2/Encryption/EncryptionInterface.php
```diff
@@ -4,8 +4,9 @@
 
 interface EncryptionInterface
 {
+    public function getSupportedAlgorithms($type = null);
     public function encode($payload, $key, $algorithm = null);
-    public function decode($payload, $key, $algorithm = null);
+    public function decode($payload, $key, $allowed_algorithms = array());
```
bshaffer: I would prefer to keep this `null` to preserve backwards compatibility, and we can change it to `array()` in the next major version

bshaffer:
Everything looks great except for the [interface change](https://github.com/bshaffer/oauth2-server-php/pull/567/files#diff-3075299424cf95cb97d7e430b4fe9f2dL8) - if we can just rename the variable but keep the method signature the same, we can avoid having to create a new major version for this (Semantic Versioning requires a new version for B.C. breaks). This will benefit everyone, and make upgrading to the security patch seamless.

src/OAuth2/Encryption/Jwt.php
phindmarsh: Yeah I wasn't exactly thrilled with the implementation either, but it seemed like the quickest way.

src/OAuth2/Encryption/EncryptionInterface.php
phindmarsh: Yep totally, forgot about that, good call.

src/OAuth2/Encryption/Jwt.php
phindmarsh: Basically yeah, it is a copy of what the `firebase/php-jwt` lib does. I realise it is an increase in complexity, and if you'd rather remove it thats all good. I just figured it wouldn't be a good idea to hardcode the supported RSA algorithms and have to maintain them separately from the original switch statement that was there. Since the different function have very similar variable signatures it's not too bad.

phindmarsh:
Replied inline to a few bits, I've updated the interface for the `decode()` method, had totally forgotten about the semver requirements. I see you've done the same already (after I'd pushed it) but it all looks good to me :+1:

phindmarsh:
One point on the semver requirements, does adding a new method to `EncryptionInterface` technically break BC too? Anyone updating would have to edit their classes implementing it to add the new `getSupportedMethods()` method. Perhaps it would be safer in [JWTBearer:180](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/GrantType/JwtBearer.php#L180) to just do something like: 

```php
$allowed = array('RS256', 'RS384', 'RS512'); 
// Verify the JWT
if (!$this->jwtUtil->decode($undecodedJWT, $key, $allowed)) {
    $response->setError(400, 'invalid_grant', "JWT failed signature verification");
    return null;
}
```

Then get rid of all that `JWT::$supportedMethods` business and deal with that in a major version bump. The downside being if the algorithms change in the `JWT` class we'd need to remember to edit them in the `JWTBearer` class too. If we declare them statically on `JWT` but not use a method to access that works too, but then creates a hard dependency where someone might be overloading the default JWT class and we've defined algorithms they don't support. Tricky business!

bshaffer:
@phindmarsh good catch! We should just remove the method from the interface, but keep everything else the same. And in the next major release, I'll incorporate `firebase/jwt-php`

phindmarsh:
If we just remove the method from the interface but someone provides an alternative implementation of `EncryptionInterface` that'll bust it too since the method will be undefined right?

Perhaps I should make a `'allowed_jwt_bearer_algorithms'` config option that has the defaults set in it (`array('RS256', 'RS384', 'RS512')`), that way we don't need anything fancy in the interface, it'll still allow people to override it, and it wont break anything backwards.

Thoughts?

phindmarsh:
Actually, technically this PR will break BC regardless because if someone isn't passing an array of allowed algorithms to `decode` it will always return false, breaking what was working for them in the past. Decode will no longer work *at all* without it. There's no real safe way to provide defaults here either. I suppose it depends on how rigid the backwards compatibility needs to be.

We _could_ fall back to the previous behaviour of trusting the header value if the supplied argument is not an array, but that's a little dodgy. If you'd prefer to keep this a minor version bump and maintain the BC in this regard then I'm happy to implement it this way.

phindmarsh:
Implemented in #569.

bshaffer:
I will be tagging a new version for this as soon as we merge #571

PR #569, opened 2015-04-22
Alternative implementation of JWT lib

phindmarsh:
An alternate patch to PR #567 that maintains backwards compatibility with existing interfaces, but is unsafe for users who call `JWT::decode` themselves without passing the allowed arguments array.

The allowed algorithms for the `JWTBearer` grant type are specified as an additional argument to the constructor, with defaults set if the argument is null.

Addresses #564.

### DIFF
--- a/src/OAuth2/GrantType/JwtBearer.php
+++ b/src/OAuth2/GrantType/JwtBearer.php
@@ -25,15 +25,17 @@ class JwtBearer implements GrantTypeInterface, ClientAssertionTypeInterface
     protected $storage;
     protected $audience;
     protected $jwtUtil;
+    protected $allowedAlgorithms;
 
     /**
      * Creates an instance of the JWT bearer grant type.
      *
-     * @param OAuth2\Storage\JWTBearerInterface $storage  A valid storage interface that implements storage hooks for the JWT bearer grant type.
-     * @param string                            $audience The audience to validate the token against. This is usually the full URI of the OAuth token requests endpoint.
-     * @param OAuth2\Encryption\JWT             $jwtUtil  OPTONAL The class used to decode, encode and verify JWTs.
+     * @param OAuth2\Storage\JWTBearerInterface|JwtBearerInterface $storage A valid storage interface that implements storage hooks for the JWT bearer grant type.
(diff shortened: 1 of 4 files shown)


src/OAuth2/GrantType/JwtBearer.php
```diff
@@ -25,15 +25,17 @@ class JwtBearer implements GrantTypeInterface, ClientAssertionTypeInterface
...
-     * @param OAuth2\Storage\JWTBearerInterface $storage  A valid storage interface that implements storage hooks for the JWT bearer grant type.
-     * @param string                            $audience The audience to validate the token against. This is usually the full URI of the OAuth token requests endpoint.
-     * @param OAuth2\Encryption\JWT             $jwtUtil  OPTONAL The class used to decode, encode and verify JWTs.
+     * @param OAuth2\Storage\JWTBearerInterface|JwtBearerInterface $storage A valid storage interface that implements storage hooks for the JWT bearer grant type.
+     * @param string $audience The audience to validate the token against. This is usually the full URI of the OAuth token requests endpoint.
+     * @param EncryptionInterface|OAuth2\Encryption\JWT $jwtUtil OPTONAL The class used to decode, encode and verify JWTs.
+     * @param null $allowedAlgorithms
      */
-    public function __construct(JwtBearerInterface $storage, $audience, EncryptionInterface $jwtUtil = null)
+    public function __construct(JwtBearerInterface $storage, $audience, EncryptionInterface $jwtUtil = null, $allowedAlgorithms = null)
```
bshaffer: This is great!! I think the only thing I would change is make the constructor the following for consistency:

```php
public function __construct(JwtBearerInterface $storage, $audience, EncryptionInterface $jwtUtil = null, $config = array())
{
    // ...
    $config = array_merge(array(
        'allowed_algorithms' => array('RS256', 'RS384', 'RS512')
    ));
    // ...
}
```
This just keeps it consistent with the rest of the library

phindmarsh: Ah yep, nice catch. Have submitted an updated commit now.

phindmarsh:
I've updated the constructor, so this should be all good to go now :+1:

bshaffer:
woohoo! Looks great!

PR #571, opened 2015-04-23
Added unset_refresh_token_after_use option

Jacketbg:
It is set to true by default to preserve the old behavior, but if set to false, it will no longer force the refresh token to expire immediately after use.

### DIFF
--- a/src/OAuth2/GrantType/RefreshToken.php
+++ b/src/OAuth2/GrantType/RefreshToken.php
@@ -24,13 +24,15 @@ class RefreshToken implements GrantTypeInterface
      *                                                      <code>
      *                                                      $config = array(
      *                                                      'always_issue_new_refresh_token' => true, // whether to issue a new refresh token upon successful token request
+     *                                                      'unset_refresh_token_after_use' => true // whether to unset the refresh token after after using
      *                                                      );
      *                                                      </code>
      */
     public function __construct(RefreshTokenInterface $storage, $config = array())
     {
         $this->config = array_merge(array(
-            'always_issue_new_refresh_token' => false
+            'always_issue_new_refresh_token' => false,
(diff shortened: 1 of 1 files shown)


bshaffer:
wow, nice work! All I ask is you submit a test for this functionality by copying one of the methods for the [Refresh Token Tests](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/OAuth2/GrantType/RefreshTokenTest.php#L72) and change the `unset_refresh_token_after_use` to `false`. Also, we'll want to add the new `config` value [here](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Server.php#L124) and [here](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Server.php#L607)

bshaffer:
added the above in cd648bc733ec31b90098d4cca2f1218720536b66

Jacketbg:
I'm relatively new to unit testing, but I will try to do something. Should I also add the new config value to the two mentioned files?

bshaffer:
No worries, I took care of it :)

Jacketbg:
That's a relief :+1: Thank you for the fast response!

pjcdawkins:
I'd just like to note here: `true` was not the previous behaviour UNLESS `always_issue_new_refresh_token` was also `true`.

So this PR actually introduced a change (and it happened to cause our implementation to break).

bshaffer:
@pjcdawkins I see what you mean. We could create a patch for this (`1.7.1`) to remove the BC-breaking change

janedbal:
I know this is old PR, but it is very ugly BC break. This should be at least mentioned in release notes...

guillerodriguez:
Looks like commit 7731601 was merged but c4330ef was not. Is that intentional? See also #824

PR #580, opened 2015-05-04
fixes BC issue from #571

bshaffer:
fixes issue noticed by @pjcdawkins


pjcdawkins:
+1!

bshaffer:
@pjcdawkins  I actually changed my mind on how to implement this - what do you think?

pjcdawkins:
Better: more explicit and each option now does what it says it will

guillerodriguez:
@bshaffer The Server class provides default values for both `always_issue_new_refresh_token` and `unset_refresh_token_after_use`. So even if user code only defines `always_issue_new_refresh_token` (see issue #824), the "isset" test in commit 7731601 will not catch this case (both config parameters will always be defined).

PR #582, opened 2015-05-08
Remove native JWT implementation and use firebase/php-jwt instead.

### LINKED ISSUE #579: Consider switching to firebase/php-jwt
When I wrote the code for JWT bearer tokens 2 years ago, there weren't many php jwt libraries around. The implementation of JWT used here was essentially my [F21/JWT](https://github.com/F21/jwt) library copied in.

Recently, it has been brought to my attention that the library suffers from a vulnerability if the same key is used for asymmetric and symmetric signing (which has since been patched). It's good to see that the issue has also been fixed here.

Having said that, I propose that we switch to [firebase/php-jwt](https://github.com/firebase/php-jwt), because they have implemented a lot of features and fixes and is the de-facto and most up-to-date jwt library for PHP.

F21:
This PR removes the native JWT implementation and uses firebase/php-jwt instead.

Resolves #579.

### DIFF
--- a/src/OAuth2/Encryption/EncryptionInterface.php
+++ b/src/OAuth2/Encryption/EncryptionInterface.php
@@ -4,8 +4,8 @@ namespace OAuth2\Encryption;
 
 interface EncryptionInterface
 {
-    public function encode($payload, $key, $algorithm = null);
-    public function decode($payload, $key, $algorithm = null);
+    public function encode($payload, $key, $algorithm = null, $keyId = null);
+    public function decode($payload, $key, $allowedAlgorithms = null);
     public function urlSafeB64Encode($data);
     public function urlSafeB64Decode($b64);
 }
--- a/src/OAuth2/Encryption/Jwt.php
+++ b/src/OAuth2/Encryption/Jwt.php
@@ -3,170 +3,45 @@
 namespace OAuth2\Encryption;
 
 /**
- * @link https://github.com/F21/jwt
- * @author F21
+ * Bridge file to use the firebase/php-jwt package for JWT encoding and decoding.
+ * @author Francis Chuang <francis.chuang@gmail.com>
  */
 class Jwt implements EncryptionInterface
 {
-    public function encode($payload, $key, $algo = 'HS256')
+    public function __construct()
     {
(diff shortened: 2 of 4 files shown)


src/OAuth2/Encryption/EncryptionInterface.php
```diff
@@ -4,8 +4,8 @@
 
 interface EncryptionInterface
 {
-    public function encode($payload, $key, $algorithm = null);
-    public function decode($payload, $key, $algorithm = null);
+    public function encode($payload, $key, $algorithm = null, $keyId = null);
+    public function decode($payload, $key, $allowedAlgorithms = array());
```
bshaffer: I don't see the win in breaking BC by changing the signature here... we should keep this as `null`

bshaffer:
This is great stuff! Two things..

1. We need to preserve backwards compatibility for the `develop` branch. However, it would be great to make this available in the `1.x` versions of this library, so I suggest moving this logic into the class `FirebaseJwt`, so it can be implemented by those who prefer it instead.

2. The changes here (which I really like) should be submitted to the [v2.x](https://github.com/bshaffer/oauth2-server-php/tree/v2.x) branch, since they break BC. Even for this branch, I would like to minimize backwards compatibility breaks however, so my interface suggestions above still apply.

src/OAuth2/Encryption/Jwt.php
```diff
@@ -3,170 +3,39 @@
...
-        if (count($tks) != 3) {
-            return false;
-        }
-
-        list($headb64, $payloadb64, $cryptob64) = $tks;
-
-        if (null === ($header = json_decode($this->urlSafeB64Decode($headb64), true))) {
-            return false;
+        if (!class_exists('\JWT')) {
+            throw new \ErrorException('You do not have firebase/php-jwt package installed. Please install it to use JWT features.');
```
bshaffer: error message should be `firebase/php-jwt must be installed to use this feature. You can do this by running "composer require firebase/php-jwt"`

F21:
See #584 and #585 instead.

bshaffer:
Yes!! Nice work.

PR #584, opened 2015-05-10
Remove native JWT implementation and use firebase/php-jwt instead.

F21:
Added some compatibility code to maintain BC.

This is for `2.0` which should be using `firebase/php-jwt` going forward.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     "suggest": {
         "predis/predis": "Required to use the Redis storage engine",
         "thobbs/phpcassa": "Required to use the Cassandra storage engine",
-        "aws/aws-sdk-php": "Required to use the DynamoDB storage engine"
+        "aws/aws-sdk-php": "Required to use the DynamoDB storage engine",
+        "firebase/php-jwt": "Required to use JWT features"
     }
 }
--- a/src/OAuth2/Encryption/EncryptionInterface.php
+++ b/src/OAuth2/Encryption/EncryptionInterface.php
@@ -4,8 +4,9 @@ namespace OAuth2\Encryption;
 
 interface EncryptionInterface
 {
-    public function encode($payload, $key, $algorithm = null);
-    public function decode($payload, $key, $algorithm = null);
+    public function encode($payload, $key, $algorithm = null, $keyId = null);
+
+    public function decode($payload, $key, array $allowedAlgorithms = array());
     public function urlSafeB64Encode($data);
     public function urlSafeB64Decode($b64);
 }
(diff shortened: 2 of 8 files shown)


src/OAuth2/Encryption/EncryptionInterface.php
```diff
@@ -4,8 +4,8 @@
 
 interface EncryptionInterface
 {
-    public function encode($payload, $key, $algorithm = null);
-    public function decode($payload, $key, $algorithm = null);
+    public function encode($payload, $key, $algorithm = null, $keyId = null);
```
bshaffer: Can we get an explanation of what the `$keyId` is, and why it needs to be passed to the `encode` function?

F21: They `keyId` is a new feature to selecting a key from a list to match against.

see https://github.com/firebase/php-jwt/blob/master/Authentication/JWT.php#L119 and https://github.com/firebase/php-jwt/blob/master/Authentication/JWT.php#L119

F21:
I have updated the PR to make use of exceptions from `firebase/php-jwt`.

bshaffer:
@F21 I hate to do this to you (and will make the change myself if you do not want to), but the new model is to merge from `develop` into `v2.x` and then remove the BC layer from there. So once we merge up to v2.x in #635, we will want to rebase this on top of it (remove `FirebaseJwt`, etc).

composer.json
```diff
@@ -21,6 +21,7 @@
     "suggest": {
         "predis/predis": "Required to use the Redis storage engine",
         "thobbs/phpcassa": "Required to use the Cassandra storage engine",
-        "aws/aws-sdk-php": "Required to use the DynamoDB storage engine"
+        "aws/aws-sdk-php": "Required to use the DynamoDB storage engine",
+        "firebase/php-jwt": "Required to use JWT features"
```
bshaffer: for v2.x I think we should just make this required instead of suggesting it, since it's core functionality (as opposed to an optional storage class)

test/OAuth2/Encryption/JwtTest.php
```diff
@@ -45,28 +45,26 @@ public function testJwtUtil($client_id, $client_key)
...
+     * @expectedException \UnexpectedValueException
+     */
     public function testInvalidJwt()
     {
         $jwtUtil = new Jwt();
 
-        $this->assertFalse($jwtUtil->decode('goob'));
-        $this->assertFalse($jwtUtil->decode('go.o.b'));
+        $jwtUtil->decode('goob');
+        $jwtUtil->decode('go.o.b');
```
bshaffer: this second line requires another test, as the first line will throw an exception and therefore halt execution

F21:
Just noticed this very stale PR. Is this safe to close @bshaffer ?

F21:
Closing this as the PR is more than 5 years old and probably irrelevant at this point.

PR #585, opened 2015-05-10
Added FirebaseJwt bridge for 1.x

F21:
Here's the PR to allow `firebase/php-jwt` to be (optionally) used with 1.x if desired.

### DIFF
--- a/src/OAuth2/Encryption/FirebaseJwt.php
+++ b/src/OAuth2/Encryption/FirebaseJwt.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace OAuth2\Encryption;
+
+/**
+ * Bridge file to use the firebase/php-jwt package for JWT encoding and decoding.
+ * @author Francis Chuang <francis.chuang@gmail.com>
+ */
+class FirebaseJwt implements EncryptionInterface
+{
+    public function __construct()
+    {
+        if (!class_exists('\JWT')) {
+            throw new \ErrorException('firebase/php-jwt must be installed to use this feature. You can do this by running "composer require firebase/php-jwt"');
+        }
+    }
+
+    public function encode($payload, $key, $alg = 'HS256', $keyId = null)
+    {
+        return \JWT::encode($payload, $key, $alg, $keyId);
+    }
+
+    public function decode($jwt, $key = null, $allowedAlgorithms = null)
+    {
+        try {
+
+            //Maintain BC: Do not verify if no algorithms are passed in.
+            if (!$allowedAlgorithms) {
+                $key = null;
+            }
+
+            return (array)\JWT::decode($jwt, $key, $allowedAlgorithms);
+        } catch (\Exception $e) {
+            return false;
+        }
+    }
+
+    public function urlSafeB64Encode($data)
+    {
+        return \JWT::urlsafeB64Encode($data);
+    }
+
+    public function urlSafeB64Decode($b64)
+    {
+        return \JWT::urlsafeB64Decode($b64);
+    }
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,5 +19,6 @@ before_script:
 - composer require predis/predis:dev-master
 - composer require thobbs/phpcassa:dev-master
 - composer require aws/aws-sdk-php:dev-master
+- composer require firebase/php-jwt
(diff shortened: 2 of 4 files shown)


src/OAuth2/Encryption/FirebaseJwt.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        try {
+
+            //Maintain BC: Do not verify if no algorithms are passed in.
+            if (!$allowedAlgorithms) {
+                $key = null;
+            }
+
+            return (array)\JWT::decode($jwt, $key, $allowedAlgorithms);
+        } catch (\Exception $e) {
```
bshaffer: What kinds of exceptions are thrown here? Can we get a comment or some explanation why this try-catch exists?

src/OAuth2/Encryption/FirebaseJwt.php
```diff
@@ -0,0 +1,47 @@
...
+    public function encode($payload, $key, $alg = 'HS256', $keyId = null)
+    {
+        return \JWT::encode($payload, $key, $alg, $keyId);
+    }
+
+    public function decode($jwt, $key = null, $allowedAlgorithms = null)
+    {
+        try {
+
+            //Maintain BC: Do not verify if no algorithms are passed in.
```
bshaffer: We don't need to maintain BC here, since this class is new.

bshaffer:
Great stuff for this PR and #584!! Thank you, @F21 

I have a few questions for clarification, otherwise :+1:

src/OAuth2/Encryption/FirebaseJwt.php
F21: This is to maintain compatibility with the rest of the code. The original JWT library returned `false` when something goes wrong (`bad jwt`, `signature not verifying`, etc). `firebase/php-jwt` throws exceptions, so we need to convert them to `false` to ensure things continue working.

src/OAuth2/Encryption/FirebaseJwt.php
F21: If that's the case, then we will need to change the default value of `$allowedAlgorithms` to `array()` and break the interface. Another issue is due to the design of the original JWT library and the recent work to fix the security issue, the last parameter `$allowedAlgorithms` is conflated to provide a list of supported algorithms or to control whether the JWT should be verified or not. For `firebase/php-jwt`, the presence of the `$key` value controls whether the JWT should be verified or not.

bshaffer: hmm, I don't think you understand my point. We cannot break BC with interfaces, but we are not required to "preserve" BC for new classes.

This function is called [here](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/GrantType/JwtBearer.php#L89) and [here])https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/GrantType/JwtBearer.php#L188), and neither instance requires this "BC" logic.



So, This

src/OAuth2/Encryption/FirebaseJwt.php
bshaffer: I guess my real question was about how `php-jwt` employs exceptions. They   list all the possible exceptions [here](https://github.com/firebase/php-jwt/blob/master/Authentication/JWT.php#L34), which is nice. 

In `2.x` it would be great to catch these at the `Response` level, so the error messages can be added to the response.

src/OAuth2/Encryption/FirebaseJwt.php
F21: In `firebase/php-jwt` `allowedAlgorithms` should not be anything but an array (and it doesn't make sense for it to be anything but an array). In this case, I believe that we should update the interface to force `allowedAlgorithms` to be an array: `array $allowedAlogrithms = array()`. This allows us to remove the code to maintain BC and also provides type hinting in IDEs.

src/OAuth2/Encryption/FirebaseJwt.php
F21: That's a good point! Will get this updated.

F21:
@bshaffer Any more feedback for this and #584?

Issue #345, opened 2014-03-07
Revoking Access / Refresh Tokens


flx5:
Hello,
I just got a bit confused about how I should handle the revokation of access tokens.
I just want to show my users the Applications they have authorized and then I want to gove them the possibility to revoke the access token.
So, my first question is how to display all applications a user is using because I couldn't find a matching function in the Storage interfaces (=> Direct MySQL Query?)
Secondly, have I to expire the Access token AND the Refresh token or only one of them?

I also must mention that I am still using the v0.9 as the System Administrator of my school just doesn't have the time to update the server.

farooquiyasir:
I think there should be a proper way of doing this. There is no relationship between an access token and its refresh token. So you cannot delete a refresh token when revoking an access token.

farooquiyasir:
I think it makes sense to revoke access of a client and not an individual access token. So if you are allowing your users to revoke access to the resource server, you will list individual clients in front of them, and will not want to repeat client names if there are more than one access tokens for a client. In that case a simple method can be written to delete / expire all access tokens + refresh token against an individual client.

phindmarsh:
Theres actually an RFC for OAuth token revocation [RFC7009](https://tools.ietf.org/html/rfc7009), which outlines some ideas on how revoking tokens should work from an API perspective.

I've implemented some changes according to this spec in my fork, if you think it might go some way to solving your use-case (and you'd like it in the lib @bshaffer) I'll he happy to issue a PR. It essentially creates a `removeAccessToken()` method on the storage interfaces and adds a revoke endpoint to the controllers. 

I suspect due to the new `removeAccessToken` method this change would probably have to land in a 2.0 release since it is not backwards compatible. Thoughts?

bshaffer:
I'd say you're exactly right on all of your points. This would be an
amazing addition to the library, and is in line with exactly what I was
hoping to add at some point, so your PR would be much appreciated!

PR #586, opened 2015-05-11
Added support for revoking tokens

phindmarsh:
Implementation follows the Draft [RFC7009](https://tools.ietf.org/html/rfc7009) OAuth 2.0 Token Revocation.

Essentially adds a unsetAccessToken method to all the storage objects, and exposes a revoke endpoint on controllers. The only interesting thing is the `ResponseType\AccessToken::revokeToken()` method which has to check all token types if the specified type doesn't exist.

This goes some of the way to addressing #345, (adds `unsetAccessToken` methods etc), but doesn't account for bulk revoking of all tokens against a client or user. For this I've just extended the built-in storage class in our implementation and added the appropriate queries for doing the bulk stuff myself.

@bshaffer I'm not sure the tests I've done are comprehensive enough, can you cast your eye over and let me know if you'd like more in a certain area.

### DIFF
--- a/src/OAuth2/ResponseType/AccessToken.php
+++ b/src/OAuth2/ResponseType/AccessToken.php
@@ -154,4 +154,41 @@ class AccessToken implements AccessTokenInterface
     {
         return $this->generateAccessToken(); // let's reuse the same scheme for token generation
     }
+
+    /**
+     * Handle the revoking of refresh tokens, and access tokens if supported / desirable
+     * RFC7009 specifies that "If the server is unable to locate the token using
+     * the given hint, it MUST extend its search across all of its supported token types"
+     *
+     * @param $token
+     * @param null $tokenTypeHint
+     * @return boolean
+     */
+    public function revokeToken($token, $tokenTypeHint = null)
+    {
+        if ($tokenTypeHint == 'refresh_token') {
+            if ($this->refreshStorage && $revoked = $this->refreshStorage->unsetRefreshToken($token)) {
+                return true;
+            }
+        }
+
+        /** @TODO remove in v2 */
+        if (!method_exists($this->tokenStorage, 'unsetAccessToken')) {
+            throw new \RuntimeException(
+                sprintf('Token storage %s must implement unsetAccessToken method', get_class($this->tokenStorage)
+            ));
+        }
+
+        $revoked = $this->tokenStorage->unsetAccessToken($token);
+
+        // if a typehint is supplied and fails, try other storages 
+        // @see https://tools.ietf.org/html/rfc7009#section-2.1
+        if (!$revoked && $tokenTypeHint != 'refresh_token') {
+            if ($this->refreshStorage) {
+                $revoked = $this->refreshStorage->unsetRefreshToken($token);
+            }
+        }
+
+        return $revoked;
+    }
 }
--- a/src/OAuth2/ResponseType/AccessTokenInterface.php
+++ b/src/OAuth2/ResponseType/AccessTokenInterface.php
@@ -20,4 +20,15 @@ interface AccessTokenInterface extends ResponseTypeInterface
      * @ingroup oauth2_section_5
      */
     public function createAccessToken($client_id, $user_id, $scope = null, $includeRefreshToken = true);
+
+    /**
+     * Handle the revoking of refresh tokens, and access tokens if supported / desirable
+     *
+     * @param $token
+     * @param $tokenTypeHint
+     * @return mixed
+     *
+     * @todo v2.0 include this method in interface. Omitted to maintain BC in v1.x
+     */
+    //public function revokeToken($token, $tokenTypeHint);
 }
--- a/test/OAuth2/Storage/AccessTokenTest.php
+++ b/test/OAuth2/Storage/AccessTokenTest.php
@@ -54,4 +54,29 @@ class AccessTokenTest extends BaseTest
         $success = $storage->setAccessToken('newtoken', 'client ID', 'SOMEOTHERID', $expires, '');
         $this->assertTrue($success);
     }
+
+    /** @dataProvider provideStorage */
+    public function testUnsetAccessToken(AccessTokenInterface $storage)
+    {
+        if ($storage instanceof NullStorage || !method_exists($storage, 'unsetAccessToken')) {
+            $this->markTestSkipped('Skipped Storage: ' . $storage->getMessage());
+
+            return;
+        }
+
+        // assert token we are unset does not exist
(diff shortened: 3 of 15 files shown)


bshaffer:
wow, this is fantastic! Your testing is great and the implementation is spot on.  I'll dive deeper in this tomorrow, but just glancing at it I have only two minor suggestions:

1. run [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) to fix some of the whitespace
2. we shouldn't add the interface changes to the `develop` branch, because it will break BC. And then we can submit the interface changes to just the `2.x` branch

phindmarsh:
No problem, what level do you run the linter at? I've just done a pass using the defaults and ended up with 92 modified files :open_mouth: 

Once you've had a chance to review and I've done the `cs-fixer` I'll close and resubmit against the `v2.x` branch

phindmarsh:
I've re-run the `php-cs-fixer` with `--level=psr2` and discarded the changes to files this PR hasn't touched - I figured you'd rather this PR was only dealing with the revoke changes, not style changes to files unaffected by it.

bshaffer:
good question, I usually run php-cs-fixer without `phpdoc_params`, so like so:

```
# php-cs-fixer fix . --fixers=-phpdoc_params
```

Even so, I found that you are right, and quite a few small cs issues seem to have escaped my notice :)

test/OAuth2/Storage/AccessTokenTest.php
```diff
@@ -54,4 +54,29 @@ public function testSetAccessToken(AccessTokenInterface $storage)
...
     }
+
+    /** @dataProvider provideStorage */
+    public function testRevokeAccessToken(AccessTokenInterface $storage)
+    {
+        if ($storage instanceof NullStorage) {
+            $this->markTestSkipped('Skipped Storage: ' . $storage->getMessage());
+
+            return;
+        }
```
bshaffer: Since this isn't part of the interface, it wouldn't be a bad idea to check the `unsetAccessToken` method exists, and skip this test if it doesn't. We would also want a @TODO to remove this for `2.x`

src/OAuth2/ResponseType/AccessTokenInterface.php
```diff
@@ -20,4 +20,13 @@
...
     public function createAccessToken($client_id, $user_id, $scope = null, $includeRefreshToken = true);
+
+    /**
+     * Handle the revoking of refresh tokens, and access tokens if supported / desirable
+     *
+     * @param $token
+     * @param $tokenTypeHint
+     * @return mixed
+     */
+    public function revokeToken($token, $tokenTypeHint);
```
bshaffer: We aren't allowed to do this for `develop`, as it will break BC

bshaffer:
Only one real required change here (we need to remove the `interface` addition). Otherwise it looks great!

bshaffer:
Then, I can merge this into `2.x` and we can add the interface changes in another commit.

phindmarsh:
Righto, I've commented out (but left a @todo note) on each of the new methods defined in the interfaces, added a check to make sure the `unsetAccessToken` method exists before running the test. 

Obviously while this doesn't break BC, it will cause a fatal error if someone attempts to use the revoke endpoint with a storage that doesn't implement that method, since there are no runtime checks to ensure that method exists. I can do this if you like, although not sure what it should do instead (throw an exception, fail quietly etc).

I assume at some point we'll put another PR against the 2.x branch with the interface methods added back again?

bshaffer:
I think throwing an exception would be the nicest thing to do for the
developer, as the message we define will be more clear than a fatal error

As for the interface changes, yes we should submit them to 2.x in a
separate PR

phindmarsh:
Noted, added a `\RuntimeException` to each call that was defined in that interface, and added a `@todo` to remove it when 2.0 lands. 

That should be all the changes needed.

phindmarsh:
Ping @bshaffer, did you want to do anything here?

bshaffer:
No, it looks great! However you will need to rebase it first, as we cannot merge it automatically.

src/OAuth2/ResponseType/AccessToken.php
```diff
@@ -153,4 +153,37 @@ protected function generateRefreshToken()
...
+            $revoked = $this->tokenStorage->unsetAccessToken($token);
+        }
+
+        if (!$revoked && $tokenTypeHint !== null) {
+            // expand search to all token types, skipping the one we've already tried
+            return $this->revokeToken($token, null, $tokenTypeHint);
+        }
+
+        return $revoked;
+    }
```
HugoCardenas: Hi, I'm merging your PR into my forked oauth repo as I needed it asap, and I was just writing some extra test and noticed some potential issue.

In this function, if you would pass some invalid ```$tokenTypeHint``` value in the first call (something different from "refresh_token" or "access_token") and the token you want to revoke would be stored in the tokenStorage, the expanded search for all token types would not work, as it would just retry once more always trying the refreshStorage and not finding anything.
If the token would be stored in refreshStorage, then it would succeed deleting it, out of luck.

I noticed that AccessTokenTest is not really testing AccessToken class, but instead just the storage underneath it. So I noticed this when writing some test for the AccessToken class itself, which would cover the behaviour of revokeToken function.

HugoCardenas: I noticed afterwards that the check for the correct value of token_type_hint is done in TokenController and this is actually test covered, so there is no issue from the POV of the project overall.

Still I think that a defensive programming is better, where every module/class/method is safe for itself and not relying on details of other code. Just my constructive 50 cents.

opengeek:
Any news on this being merged? I am looking for exactly this feature without resorting to directly manipulating the storage...

phindmarsh:
Sorry I keep forgetting to rebase this up to master. I'll do it now and then it should be ready to go.

phindmarsh:
Righto, all up to date now, @bshaffer this is ready when you are.

bshaffer:
These changes are great @phindmarsh! However, @HugoCardenas has a great point. I am submitting a PR to your `token-revoke` branch with a few changes to fix that issue, and once you merge that one in (and rebase out the merge commit) this'll be good to go

bshaffer:
added phindmarsh/oauth2-server-php/pull/1 to address these issues

phindmarsh:
Merged and rebased back to a single commit, thanks @bshaffer :+1:

bshaffer:
and there was much rejoicing! Thank you @phindmarsh  and everyone else involved for your hard work on this!!

PR #587, opened 2015-05-11
[v1.7.1] fixes BC issue from #571

bshaffer:


src/OAuth2/GrantType/RefreshToken.php
```diff
@@ -34,6 +34,14 @@ public function __construct(RefreshTokenInterface $storage, $config = array())
             'always_issue_new_refresh_token' => false,
             'unset_refresh_token_after_use' => true
         ), $config);
+
+        // to preserve B.C. with v1.6
+        // @see https://github.com/bshaffer/oauth2-server-php/pull/580
+        // @todo - remove in v2.0
+        if (isset($config['always_issue_new_refresh_token']) && !isset($config['unset_refresh_token_after_use'])) {
+            $this->config['unset_refresh_token_after_use'] = $config['always_issue_new_refresh_token'];
```
svycka: @bshaffer sorry but this does not fix BC and is totally wrong, how this if can be executed if just few lines above both config options are set but this condition expects one to be set and another not set while both are defined.

svycka: woops looks I got confused with $config and $this->config so this if could be ok but still I have a BC just not sure why. Investigating... Is test written to confirm this fix works?

bshaffer: This line checks the $config parameter, while the options being set above it are setting $this->config

svycka: Okay might found the reason. I am creating server like in documentation example http://bshaffer.github.io/oauth2-server-php-docs/grant-types/refresh-token/#implementation
```php
// create a storage object to hold refresh tokens
$storage = new OAuth2\Storage\Pdo(array('dsn' => 'sqlite:refreshtokens.sqlite'));

// create the grant type
$grantType = new OAuth2\GrantType\RefreshToken($storage);

// add the grant type to your OAuth server
$server->addGrantType($grantType);
```
as you can see creating `OAuth2\GrantType\RefreshToken($storage)` without $config so in this constructor $config is empty that's why this fix does not help. `always_issue_new_refresh_token` default was false so I didn't added `always_issue_new_refresh_token` here and now have problem.

svycka: @bshaffer do you think this should be fixed? or this works as expected?

bshaffer: It should probably be fixed, yes, as this inadvertently changed the default
behavior.

PR #598, opened 2015-05-26
Allow configurable access_lifetime and refresh_token_lifetime for default JWT tokens

quintonparker:
Allows the default access_lifetime and refresh_token_lifetime values to be overriden when default JWT tokens are implied in the following OAuth2\Server configuration

```
$server = new OAuth2\Server($storage, array(
    'always_issue_new_refresh_token' => true,
    'access_lifetime' => 316,
    'refresh_token_lifetime' => 499,
    'use_jwt_access_tokens' => true
));
```

Without this patch the specified access_lifetime and refresh_token_lifetime values is ignored

### DIFF
--- a/src/OAuth2/Server.php
+++ b/src/OAuth2/Server.php
@@ -693,7 +693,7 @@ class Server implements ResourceControllerInterface,
             $refreshStorage = $this->storages['refresh_token'];
         }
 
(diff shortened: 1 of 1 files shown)


bshaffer:
Good catch, thanks!

PR #610, opened 2015-06-24
Allow overrides on user identity columns avoiding BC breaks

dkcwd:
Hi @bshaffer, hope you're well and thanks for the lib.

I added a config array for the columns used in the getUser and setUser methods avoiding BC breaks.
Also added a wrapper for the default sha1 password hashing to simplify reuse.

My use case for this change is to leverage the library where I will authenticate the identity of a user where the username and password are stored in a MySQL database where 'email' is the equivalent of 'username'.

I didn't want to recreate all the basic queries by overriding methods and instead wanted to have to only override the hashPassword method via inheritance.

If this is merged I'll be more than happy to add other changes like this for other adapters.

P.S. I noticed the Travis CI build failed because the travis.yml is using:
```
- composer require aws/aws-sdk-php:dev-master
```
It passes on the test servers using >= PHP 5.5 but the aws/aws-sdk-php:dev-master requires >= PHP 5.5 so the 5.4 and 5.3 builds are broken.

See https://travis-ci.org/dkcwd/oauth2-server-php/jobs/68149259
And https://travis-ci.org/dkcwd/oauth2-server-php for the overview

It's a good time to review how worthwhile it is supporting the older versions or fixing your composer requirements on a version of the AWS lib to be one which works with earlier versions of PHP.

Here's a link to a discussion around 5.5 going into security updates only from this week:
http://news.php.net/php.internals/86784

Here's the commit where the AWS lib composer file was updated to require >= 5.5
https://github.com/aws/aws-sdk-php/commit/6335652ec049e6761c54fe3110a1523edab3c0d0

If I can help to get this PR merged or if you want to discuss changes just holler.

Cheers,
Dave

### DIFF
--- a/src/OAuth2/Storage/Pdo.php
+++ b/src/OAuth2/Storage/Pdo.php
@@ -57,17 +57,29 @@ class Pdo implements
         // debugging
         $connection->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
 
-        $this->config = array_merge(array(
-            'client_table' => 'oauth_clients',
-            'access_token_table' => 'oauth_access_tokens',
-            'refresh_token_table' => 'oauth_refresh_tokens',
-            'code_table' => 'oauth_authorization_codes',
-            'user_table' => 'oauth_users',
-            'jwt_table'  => 'oauth_jwt',
-            'jti_table'  => 'oauth_jti',
-            'scope_table'  => 'oauth_scopes',
-            'public_key_table'  => 'oauth_public_keys',
-        ), $config);
+        $this->config = array_merge(
+            array(
+                'client_table' => 'oauth_clients',
+                'access_token_table' => 'oauth_access_tokens',
+                'refresh_token_table' => 'oauth_refresh_tokens',
+                'code_table' => 'oauth_authorization_codes',
+                'user_table' => 'oauth_users',
+                'jwt_table'  => 'oauth_jwt',
+                'jti_table'  => 'oauth_jti',
+                'scope_table'  => 'oauth_scopes',
+                'public_key_table'  => 'oauth_public_keys',
+            ),
+            array(
+                // defaults for column names in user queries
+                'identity' => array(
+                    'username' => 'username',
+                    'password' => 'password',
+                    'first_name' => 'first_name',
+                    'last_name' => 'last_name',
+                )
+            ),
+            $config
+        );
     }
 
     /* OAuth2\Storage\ClientCredentialsInterface */
--- a/src/OAuth2/Storage/Pdo.php
+++ b/src/OAuth2/Storage/Pdo.php
@@ -317,7 +317,13 @@ class Pdo implements
 
     public function getUser($username)
     {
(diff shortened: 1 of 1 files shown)


dkcwd:
@bshaffer just wanted to let you know that the reason the build failed above is a fairly major issue surrounding ongoing support for PHP 5.3 and PHP 5.4 with the library.

The current develop build fails on Travis for 5.3 and 5.4 even when making even changes to documentation due to the AWS dependency. 

Here is my Travis build with only a change to the README.md https://travis-ci.org/dkcwd/oauth2-server-php/jobs/68900529

Here is the commit history for my test branch: https://github.com/dkcwd/oauth2-server-php/commits/feature/testing-current-develop-build 

Feel free to get in touch, I'd really like to have the option of using the extra config which my PR makes available.

src/OAuth2/Storage/Pdo.php
```diff
@@ -317,17 +335,39 @@ public function getUser($username)
             'user_id' => $username
         ), $userInfo);
     }
+    
+    public function hashPassword($password)
```
bshaffer: Why is this `public` (as opposed to `protected`) ?

src/OAuth2/Storage/Pdo.php
```diff
@@ -57,17 +57,29 @@ public function __construct($connection, $config = array())
...
+                'access_token_table' => 'oauth_access_tokens',
+                'refresh_token_table' => 'oauth_refresh_tokens',
+                'code_table' => 'oauth_authorization_codes',
+                'user_table' => 'oauth_users',
+                'jwt_table'  => 'oauth_jwt',
+                'jti_table'  => 'oauth_jti',
+                'scope_table'  => 'oauth_scopes',
+                'public_key_table'  => 'oauth_public_keys',
+            ),
+            array(
```
bshaffer: Why isn't this a part of the array above? It would be more straightforward to have a single default config array.

src/OAuth2/Storage/Pdo.php
dkcwd: I'm comfortable with leaving that method accessible from outside of the object but equally happy to change that if it's a blocker to moving forward with the change.

src/OAuth2/Storage/Pdo.php
dkcwd: I wanted to keep tables names and identity columns separate but interested in your opinion.
What are your thoughts based on that?

src/OAuth2/Storage/Pdo.php
```diff
@@ -57,17 +57,27 @@ public function __construct($connection, $config = array())
...
+                'access_token_table' => 'oauth_access_tokens',
+                'refresh_token_table' => 'oauth_refresh_tokens',
+                'code_table' => 'oauth_authorization_codes',
+                'user_table' => 'oauth_users',
+                'jwt_table'  => 'oauth_jwt',
+                'jti_table'  => 'oauth_jti',
+                'scope_table'  => 'oauth_scopes',
+                'public_key_table'  => 'oauth_public_keys',
+                // defaults for column names in user queries
+                'identity' => array(
```
dkcwd: Now part of the same array....

PR #612, opened 2015-06-30
Device authotization (device flow) http://tools.ietf.org/html/draft-ietf-oauth-v2-05#section-3.7

TroRg:
Hello, i don't know if i properly implement this feature, but it works and will be good to have it from the box.

### DIFF
--- a/src/OAuth2/GrantType/DeviceCode.php
+++ b/src/OAuth2/GrantType/DeviceCode.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace OAuth2\GrantType;
+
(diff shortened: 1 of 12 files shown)


bshaffer:
The Device flow was removed in favor of the implicit flow in the final Oauth2 draft. I imagine you realize this, and is why this PR is close?

Spomky:
This grant type flow was removed, but the Implicit grant type flow does not superseded it. 
Google for example still have an implementation of this grant type. It is reality helpful for apps on TV screens for example.

TroRg:
I found many mistakes in my code. Need more time to understand code and all processes. =)

PR #621, opened 2015-07-24
fix aws composer dependencies

bshaffer:
for tests running PHP <5.5


bshaffer:
closing - we need to only use `2.8` for now

PR #625, opened 2015-08-04
Added content type 'multipart/form-data' to allow the capability to u…

coachroebuck:
…pload binary data while using this tool, if desired.

### DIFF
--- a/src/OAuth2/TokenType/Bearer.php
+++ b/src/OAuth2/TokenType/Bearer.php
@@ -113,7 +113,7 @@ class Bearer implements TokenTypeInterface
                 $contentType = substr($contentType, 0, $pos);
             }
 
-            if ($contentType !== null && $contentType != 'application/x-www-form-urlencoded') {
+            if ($contentType !== null && $contentType != 'application/x-www-form-urlencoded' && $contentType != 'multipart/form-data') {
                 // IETF specifies content-type. NB: Not all webservers populate this _SERVER variable
                 // @see http://tools.ietf.org/html/rfc6750#section-2.2
                 $response->setError(400, 'invalid_request', 'The content type for POST requests must be "application/x-www-form-urlencoded"');


coachroebuck:
I've read across conflicting opinions online, most notably here:

http://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data

I also read up on the documentation related to the form-encoded body parameter:

http://tools.ietf.org/html/rfc6750#section-2.2

I have a situation in which I must upload binary data. Uploads were failing because of the line in question. The proposed solution crossed my mind. 

I decided to push this up and get feedback from the community. Perhaps someone else may have a better idea.

btw: I see there's an automated process that got triggered in response to submitting a merge request: https://travis-ci.org/bshaffer/oauth2-server-php/builds/74029210

phindmarsh:
The only trouble with adding this is the current behaviour is defined by [RFC 6750 section 2.2](http://tools.ietf.org/html/rfc6750#section-2.2), so allowing other content types is not permitted by the spec. When including the token in the token in the body:

> * The HTTP request entity-header includes the "Content-Type" header
      field set to "application/x-www-form-urlencoded".
> * The entity-body follows the encoding requirements of the
      "application/x-www-form-urlencoded" content-type as defined by
      HTML 4.01 [W3C.REC-html401-19991224].
> * The HTTP request entity-body is single-part.

The alternative to this is to supply the token as an `Authorization` header or in the query string of the URL. In these two cases, you are allowed to use whatever `Content-Type` you like. 

Is it possible for you to use the token in a header or in the query string instead? That would allow you to use whatever `Content-Type` you want and still conform to the spec.

coachroebuck:
Thanks for the feedback! Your note revealed a problem with the client: that access_token was part of the POST request. 

Mobile apps: I believe I can. I can have the mobile apps convert the binary data into something more manageable data.

Web browsers: I'll turn around and test that image upload html page tonight. 

I'll go ahead and close this merge request. In the meantime, is it OK for me to reach out to you offline in the event I continue to run into problems? (I don't think I will, but just in case).

Thanks again!

phindmarsh:
Yep no problem, I have implemented uploads in both mobile and JS web-apps so it can be done quite easily, it just depends on the libraries you are using.

PR #636, opened 2015-08-27
F21 firebase jwt 1.x

bshaffer:


bshaffer:
will resolve conflicts in #585

PR #639, opened 2015-09-04
Mongo Storage HHVM Compability

jniebuhr:
When running in hhvm passing a value as reference will result in a fatal error. This pull request makes sure everything passed to the MongoCollection::insert method is a variable.


Cyruxx:
+1

bshaffer:
@jniebuhr @Cyruxx thank you for catching this

PR #642, opened 2015-09-17
add storage CassandraCQL

tobias74:
the existing Cassandra Storage uses a library that is conflicting with the php-cassandra-driver (datastax) I am using. For exercise and fun I adapted the existing Cassandra-Stroage to work with CQL and the datastax-php-driver.

Maybe someone finds this usefull :-)

Tobias

### DIFF
--- a/src/OAuth2/Storage/CassandraCQL.php
+++ b/src/OAuth2/Storage/CassandraCQL.php
@@ -1,0 +1,448 @@
+<?php
+
+namespace OAuth2\Storage;
+
+use OAuth2\OpenID\Storage\UserClaimsInterface;
+use OAuth2\OpenID\Storage\AuthorizationCodeInterface as OpenIDAuthorizationCodeInterface;
+
+/**
+ * Cassandra storage based on CQL
+ * 
+ * This storage uses the datastax php-driver for cassandra
+ * https://github.com/datastax/php-driver
+ * 
+ * <code>
+ *  $cassandraCluster = \Cassandra::cluster()->withContactPoints('192.168.0.100')->withPort(9042)->build();
+ *  $cassandraSession = $cassandraCluster->connect( 'my_keyspace' );
+ * </code>
+ *
+ * You need the following tables in cassandra for this storage
+ * <code>
+ *   $sql = "CREATE TABLE IF NOT EXISTS ".$this->config['column_family_data']." (key varchar, data text, PRIMARY KEY(key))";
+ *   $statement = new \Cassandra\SimpleStatement($sql);
+ *   $result = $this->getSession()->execute($statement);
+ *
+ *   $sql = "CREATE TABLE IF NOT EXISTS ".$this->config['column_family_clients']." (key text, data text, PRIMARY KEY(key))";
+ *   $statement = new \Cassandra\SimpleStatement($sql);
+ *   $result = $this->getSession()->execute($statement);
+ * </code> 
+ * 
+ * Then, register the storage client:
+ * <code>
+ *  $storage = new OAuth2\Storage\Cassandra($cassandra);
+ *  $storage->setClientDetails($client_id, $client_secret, $redirect_uri);
+ * </code>
+ *
+ */
+class CassandraCQL implements AuthorizationCodeInterface,
+    AccessTokenInterface,
+    ClientCredentialsInterface,
+    UserCredentialsInterface,
+    RefreshTokenInterface,
+    JwtBearerInterface,
+    ScopeInterface,
+    PublicKeyInterface,
+    UserClaimsInterface,
+    OpenIDAuthorizationCodeInterface
+{
+
+    protected $cassandraSession;
+
+    protected $config;
+
+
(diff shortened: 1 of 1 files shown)


tobias74:
since I have copypasted lots of code from the original cassandra storage, maybe there should be a common base-class for both storages.

I did not want to be too intrusive, so I left existing code unchanged and just added the new storage.

bshaffer:
Thanks Tobias! I'm ok with not having a common base class. However we do
need to add your Storage Adapter to BaseTest and Test/Bootstrap so it can
get tested with the rest of the test suite.

bshaffer:
@tobias74 any word on this? I can provide you with more information on how to add these tests.

tobias74:
I will add the tests this weekend, I have to get up to speed with Travis-CI. :-)
How could I run the tests locally?

bshaffer:
Just pull the repo, run `composer install`, and then run `phpunit`... you should be good to go.

It'll skip tests automatically it can't connect to, but if you run `phpunit -v` you'll see the output for how to set those up.

Dylan1312:
This is exactly what I was about to start writing! Thanks @tobias74

bshaffer:
@tobias74 any word on this? Will you add this to the tests [here](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/BaseTest.php)?

bshaffer:
You will also need to add more information about the dependent Cassandra client library, and also type-hinting in your constructor.

bshaffer:
To answer your question about testing, just pull down the library and run `phpunit` in the project root. It'll skip the other storage tests unless you have them running, which is fine because you don't need to test those. Just model your tests after the [existing cassandra storage class](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/Bootstrap.php#L216)

tobias74:
I have added the tests according to the existing Cassandra-Storage. Three of my tests are still failing when I run phpunit locally:

1) Warning
The data provider specified for OAuth2\OpenID\Storage\AuthorizationCodeTest::testCreateAuthorizationCode is invalid.
line 1:173 extraneous input ')' expecting EOF

2) OAuth2\OpenID\Storage\UserClaimsTest::testGetUserClaims with data set #9 (OAuth2\Storage\CassandraCQL Object (...))
Failed asserting that 'testuser@test.com' matches expected null.

3) OAuth2\Storage\PublicKeyTest::testSetAccessToken with data set #9 (OAuth2\Storage\CassandraCQL Object (...))
Failed asserting that '-----BEGIN CERTIFICATE-----
MIICiDCCAfGgAwIBAgIBADANBgkqhkiG9w0BAQQFADA9MQswCQYDVQQGEwJVUzEL
...
FRrlM1f6s9VTLWvwGItjfrof0Ba8Uq7ZDSb9Xg==
-----END CERTIFICATE-----' matches expected null.

I am still looking into that, maybe someone could point me in the right direction?
Thank you :-)

bshaffer:
These tests are passing for the existing `OAuth2\Storage\Cassandra` implementation, so most likely you'll need to debug your added storage class to make the tests pass. If you crack open the tests themselves, you'll see where the assertions are being made. 

The first error (`extraneous input ')' expecting EOF`) seems like you have a syntax error in your CQL.
The second and third error seem to imply your Cassandra test data has not been set up correctly (see [OAuth2\Storage\Bootstrap::createCassandraDB](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/Bootstrap.php#L258))

Good luck!

bshaffer:
So a question to help my own cassandra understanding - what are the advantages / disadvantages to the two cassandra libraries being used? Is one considered deprecated, or are they both viable options? Do the names `Cassandra` and `CassandraCQL` properly describe the differences between them?

tobias74:
Yes, I had an error in my CQL! Now the tests seem to pass:

---
OK, but incomplete, skipped, or risky tests!
Tests: 407, Assertions: 1197, Skipped: 126.
---

Concerning advantages and disadvantages: I don't know :-)
I had to choose a library in order to get started with cassandra. So I went with the datastax-library because it came up first on google and the given examples used CQL-Queries. 

One downside of the datastax-library is the somewhat involved setup-process ("pecl install cassandra" throws errors on my machine, so I had to build from source manually).

The name "CassandraCQL" is just a working title, I am open to suggestions :-)

Dylan1312:
To add, phpcassa is deprecated and the author recommends moving to Datastax.

Also @tobias74, I noticed you haven't made us of Cassandra's TTL to have tokens automatically expire. The $expire variable passed into setValue() is ignored. I wrote an example quickly if you'd like to look -> https://github.com/Dylan1312/oauth2-server-php/commit/b583ae8edd9486feef4844d05f9add1a5baa1c7a

tobias74:
Thanks @Dylan1312 :-)
should I just copy your code or do you want to issue a pull-request?

Dylan1312:
Go ahead and copy if you're happy with it :+1:

PR #645, opened 2015-09-24
Allow null for client_secret

Wilt:
Changed `client_secret` declaration in schema to allow NULL values


Wilt:
I added pull requests for this.
Once in the database schema:
https://github.com/bshaffer/oauth2-server-php/pull/645
Two times in the docs:
https://github.com/bshaffer/oauth2-server-php-docs/pull/68
https://github.com/bshaffer/oauth2-server-php-docs/pull/69

PR #650, opened 2015-10-06
Added support for table prefixes to PDO

chestr92:
This is a simple but very useful feature that I think demand not only by me.

It can also be done through the environment variable, then it definitely will not break anything around. I wrote this patch for PDO and if you take it, I will be very grateful to you.

I could use a fork of your project, but then there will be further problems with its support from my side.

Thanking you in advance!


copynpaste:
What about passing in the updated table names in config (constructor) to overwrite default ones?

chestr92:
oh my... its was stupid... sorry...

PR #651, opened 2015-10-06
fix bug in isPublicClient of Cassandra-Storage (revised pull-request)

tobias74:
this is a revised pull-request concerning a small bug in the isPublicClient-Method of the Cassandra-Storage


bshaffer:
This is a good catch! And it reveals a hole in our code coverage. Thanks!

PR #654, opened 2015-10-13
Feature/fix arround oidc spec

mokemokechicken:
Fix for pass test of http://openid.net/certification/testing/

* HTTP Response of TokenEndpoint should return application/json Content-Type
* fix at_hash computation in IdToken.

### DIFF
--- a/src/OAuth2/Controller/TokenController.php
+++ b/src/OAuth2/Controller/TokenController.php
@@ -51,7 +51,8 @@ class TokenController implements TokenControllerInterface
             // server MUST disable caching in headers when tokens are involved
             $response->setStatusCode(200);
             $response->addParameters($token);
-            $response->addHttpHeaders(array('Cache-Control' => 'no-store', 'Pragma' => 'no-cache'));
+            $response->addHttpHeaders(array('Cache-Control' => 'no-store', 'Pragma' => 'no-cache',
+                'Content-Type' => 'application/json'));
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/OAuth2/OpenID/ResponseType/IdToken.php
```diff
@@ -6,6 +6,7 @@
 use OAuth2\Encryption\Jwt;
 use OAuth2\Storage\PublicKeyInterface;
 use OAuth2\OpenID\Storage\UserClaimsInterface;
+use Phalcon\Logger;
```
copynpaste: Remove this pls

mokemokechicken: I'm sorry. I deleted it!

bshaffer:
This is great! Thank you for finding this and fixing these.

src/OAuth2/Controller/TokenController.php
```diff
@@ -51,7 +51,8 @@ public function handleTokenRequest(RequestInterface $request, ResponseInterface
             // server MUST disable caching in headers when tokens are involved
             $response->setStatusCode(200);
             $response->addParameters($token);
-            $response->addHttpHeaders(array('Cache-Control' => 'no-store', 'Pragma' => 'no-cache'));
+            $response->addHttpHeaders(array('Cache-Control' => 'no-store', 'Pragma' => 'no-cache',
+                'Content-Type' => 'application/json'));
```
casperbiering: This change should not be needed, since it's set when calling send method on Reponse object.

bshaffer: How would it know what content type to use unless we set it?

casperbiering: We _do_ set it. When calling $response->send();
https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Response.php#L203

bshaffer: That's only implemented in my response class, not as part of the interface.

bshaffer:
closing in favor of #698 (rebased and whitespace cleanup)

Issue #381, opened 2014-05-09
Getting User Info upon Login


martinzaleski:
I need to get the userid when logging in, and I just can't find the way to do it.

I'm sending this info:
{
            "username":USERNAME,
            "password": PASSWORD,
            "grant_type" : "password",
        "client_id" : CLIENT

}
and I get the following response:
{
access_token: "23f32b51076912b193b648c0b54ddb518546fa38"
expires_in: 3600
token_type: "Bearer"
scope: null
refresh_token: "233086f639d78b77896b2f2a583072809c537910"
}

I just can't get the user info.

On the other hand, it would be nice as well to be able to, in random calls, return information related to the user making the call, basing on the bearer token without needing to have an ID posted by GET or POST.
i.e:
www.myapp.com/api/getcollections ---cool
www.myapp.com/api/getcollections/id ---not cool

Thank you!!

theroadrunnershow:
https://developers.google.com/drive/web/credentials

handokox8:
@bshaffer how about userinfo endpoint ? I'm looking for userinfo endpoint documentation. After implement openid with this lib, I got id_token. Then how handle this, so it can be used to access userinfo. Thanks.

PR #662, opened 2015-11-05
make access token contain user identifier

DCMNMarc:
If a user gets authenticated by a client the client should get the user identifier.

For example I have an endpoint to get user information `https://api.example.com/user/` returning a list of users as well as `https://api.example.com/user/:id` returning information of one user only, but there is no endpoint returning only the current authenticated user.

So if the client directly receives the user identifier it can call `https://api.example.com/user/:id` to get additional information and it would match the one endpoint per resource principle.


DCMNMarc:
duplicate of #381

bshaffer:
It sounds like you want to use [OpenIDConnect](http://bshaffer.github.io/oauth2-server-php-docs/overview/openid-connect/) for this.

PR #664, opened 2015-11-09
[composer] Consolidate dev dependencies

Maks3w:


.travis.yml
```diff
@@ -15,11 +20,11 @@ services:
 - mongodb
 - redis-server
 - cassandra
+before_install:
+- phpenv config-rm xdebug.ini || return 0
```
bshaffer: why is this call desirable? Have you seen problems with `xdebug` on travis?

.travis.yml
```diff
@@ -1,10 +1,15 @@
 language: php
 sudo: false
+cache:
+  directories:
+    - $HOME/.composer/cache
+    - vendor
```
bshaffer: This is great! Thanks for adding this.

.travis.yml
```diff
@@ -1,10 +1,15 @@
...
 php:
 - 5.3
 - 5.4
 - 5.5
 - 5.6
+- 7
```
bshaffer: `7` errors out, and is required to be `7.0` (now that there is an official release).

This can be removed, as it's been added in a separate PR

.travis.yml
Maks3w: HHVM does not have this file. Just silently fail ig XDebug is not available

Maks3w: Also if you don't do test coverage this speed up the execution byu 2x or 3x

bshaffer:
closing in favor of #694 (I just rebased it and submitted to the proper branch)

PR #665, opened 2015-11-09
Fix #665. Wrong validation for empty registered uris

### LINKED ISSUE #663: Two whitespaces on redirect_uri works as a wildcard
Given this redirect_uri `" http://localhost"` makes valid any `$supplied_redirect_uri` if `require_exact_redirect_uri = false`.

Because after explode/split the string by the whitespace character
https://github.com/bshaffer/oauth2-server-php/blob/40e2e32786a76463b721e84d1ee7e7959abcd7c0/src/OAuth2/Controller/AuthorizeController.php#L336

```php
$registered_uri = "";
$inputUri = "any";
var_dump(strcasecmp(substr($inputUri, 0, strlen($registered_uri)), $registered_uri) === 0); // true
```
https://github.com/bshaffer/oauth2-server-php/blob/40e2e32786a76463b721e84d1ee7e7959abcd7c0/src/OAuth2/Controller/AuthorizeController.php#L346

Is this behavior expected or should be protected against empty `$registered_uri` values?

I found this behavior can be achieved accidentally if `$redirect_uri` is constructed by implode a `null` value (implode(" ", [null, "http://localhost"]);

Maks3w:
Fix #663

### DIFF
--- a/src/OAuth2/Controller/AuthorizeController.php
+++ b/src/OAuth2/Controller/AuthorizeController.php
@@ -341,9 +341,14 @@ class AuthorizeController implements AuthorizeControllerInterface
                     return true;
                 }
             } else {
+                $registered_uri_length = strlen($registered_uri);
+                if ($registered_uri_length === 0) {
+                    return false;
+                }
+
                 // the input uri is validated against the registered uri using case-insensitive match of the initial string
                 // i.e. additional query parameters may be applied
-                if (strcasecmp(substr($inputUri, 0, strlen($registered_uri)), $registered_uri) === 0) {
+                if (strcasecmp(substr($inputUri, 0, $registered_uri_length), $registered_uri) === 0) {
                     return true;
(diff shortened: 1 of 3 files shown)


src/OAuth2/Controller/AuthorizeController.php
```diff
@@ -341,9 +341,14 @@ protected function validateRedirectUri($inputUri, $registeredUriString)
                     return true;
                 }
             } else {
+                $registered_uri_length = strlen($registered_uri);
+                if ($registered_uri_length === 0) {
```
bshaffer: would be cleaner: `if (empty($registered_uri)) {`

Maks3w: micro optimization by using the same result in both statements.

bshaffer:
closing in favor of #695

PR #667, opened 2015-11-12
Don't return claims with `null` value

afilippov1985:
According to [5.3.2. Successful UserInfo Response](http://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse)
If a Claim is not returned, that Claim Name SHOULD be omitted from the JSON object representing the Claims; it SHOULD NOT be present with a null or empty string value.

However, I believe that returning empty string is OK.
What do you think?

### DIFF
--- a/src/OAuth2/OpenID/Storage/UserClaimsInterface.php
+++ b/src/OAuth2/OpenID/Storage/UserClaimsInterface.php
@@ -18,7 +18,7 @@ interface UserClaimsInterface
     const ADDRESS_CLAIM_VALUES  = 'formatted street_address locality region postal_code country';
     const PHONE_CLAIM_VALUES    = 'phone_number phone_number_verified';
 
-    const CLAIM_EMAIL_NUMBER_VERIFIED = 'email_verified';
+    const CLAIM_EMAIL_VERIFIED = 'email_verified';
     const CLAIM_PHONE_NUMBER_VERIFIED = 'phone_number_verified';
 
     /**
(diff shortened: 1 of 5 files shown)


src/OAuth2/OpenID/Storage/UserClaimsInterface.php
```diff
@@ -9,7 +9,8 @@
 interface UserClaimsInterface
 {
     // valid scope values to pass into the user claims API call
-    const VALID_CLAIMS = 'profile email address phone';
+    const VALID_SCOPE_VALUES = 'profile email address phone';
```
bshaffer: Unfortunately, we can't remove a class constant w/o breaking BC, so this has to be named back to `VALID_CLAIMS`

src/OAuth2/Storage/Cassandra.php
```diff
@@ -423,43 +423,54 @@ public function getEncryptionAlgorithm($client_id = null)
...
+
+            if (count($addressClaims)) {
+                $userClaims[self::SCOPE_ADDRESS] = $addressClaims;
+            }
+        }
+        else {
+            foreach ($claimValues as $claimValue) {
+                if (isset($userDetails[$claimValue])) {
+                    if (in_array($claimValue, array(self::CLAIM_EMAIL_VERIFIED, self::CLAIM_PHONE_NUMBER_VERIFIED), true)) {
+                        $userDetails[$claimValue] = (is_string($userDetails[$claimValue]) && !strcasecmp($userDetails[$claimValue], 'true')) || ((is_numeric($userDetails[$claimValue]) || is_bool($userDetails[$claimValue])) && (bool)$userDetails[$claimValue]);
```
bshaffer: Please break this out across multiple lines and add comments, so the logic is clear.

bshaffer:
An empty string seems okay to me, I agree. I'd be curious what the reasoning is there.

We need tests for this, and also this code is very difficult to follow.

PR #672, opened 2015-11-18
Implemented method to override the password hashing algorithm

ashiina:
By having `sha1($password)` hard-coded in the storage classes, it lacks flexibility for users who want to select a different password hashing algorithm. 
I encapsulated the hashing algorithm in an overridable method `hashPassword()`.


bshaffer:
LGTM

PR #676, opened 2015-12-04
Suggestion to create base class for key-value storages

afilippov1985:
Adding another storage will be as easy as to define few functions
`get`, `set`, `delete` and possibly `serialize`/`unserialize`.


bshaffer:
a few thoughts on this:

1. `get`, `set`, etc should be protected. 
2. The base class is only for one storage class
3. There are enough subtle differences between storage engines, I'm not convinced this will be desirable

bshaffer:
This is looking a lot better!

afilippov1985:
Further work blocked by issue #677

afilippov1985:
[mongo](http://php.net/manual/en/mongo.manual.php) extension is deprecated.
So, I wrote other storage adapter for MongoDB using new extension [mongodb](http://php.net/manual/en/set.mongodb.php)

PR #690, opened 2016-01-02
Fixing various bugs with JWT part

mtangoo:


bshaffer:
closing this, as it is wrong and there is really not anything good about it. Sorry :(

Issue #675, opened 2015-12-04
Inconsistency in storage interfaces


afilippov1985:
Somewhere function must return `false` if value not exist (`ClientInterface::getClientDetails()`)
Somewhere function must return `null` in such case (`AuthorizationCodeInterface::getAuthorizationCode()`)
Somewhere behavior not defined (`PublicKeyInterface::getPrivateKey()`)

I suggest to revise interfaces and
return `null` if value not exist (not defined)
return `false` on error

bshaffer:
This seems like a good idea. I seem to remember returning `false` for a specific reason, but it may have just been because `Pdo` returns false from its queries, which is not a real good reason.

afilippov1985:
Also, existing storages has functions to set data for unit tests (such as `setClientDetails`, `setScope`, `setClientKey`, `unsetAccessToken`, `setUser`). They should be added to corresponding storage interfaces.

bshaffer:
I disagree with this second request. These are convenience functions, and if you add your own storage classes, you are free to add those functions or not add those functions.  The library classes require them for testing, but the actual OAuth server does not require them.

PR #697, opened 2016-01-12
Suggestion to create base class for key-value storages 2

afilippov1985:
It is renewed PR #676.

Rewritten storages:
* Cassandra
* CouchbaseDB
* Redis

Added storages:
* Memcached
* MongoDB


copynpaste:
I like this approach. Could you add some very descriptive doc-blocks in that change pls. :smile:

bshaffer:
This is awesome! a few things
 - we need to convert the tabs to spaces
 - have you considered using traits instead of a base class?

afilippov1985:
Hmm.. I don't see any tabs.
Unfortunately, traits can't implement interfaces.

afilippov1985:
though... storage class can declare that it implements interfaces, but actual implementation will be done by trait.
Then ...  would be better to create several traits (one for each interface), each will implement corresponding interface.

bshaffer:
Yes, that is what I had in mind! This is tricky though because this would
break compatibility with PHP 5.3.

afilippov1985:
Yes, I also wanted to mention about this. Traits require PHP >=5.4

afilippov1985:
Oops...  travis not working [https://travis-ci.org/bshaffer/oauth2-server-php/jobs/102334251]

afilippov1985:
Waiting for PR #701

bluebaroncanada:
So, with #701, this is passing?

afilippov1985:
Rechecked.
Yes, it passes tests.
Current `develop` branch also passes tests with #701

bluebaroncanada:
So @bshaffer, unless you say otherwise, I'll merge this and #701 in about a week.

src/OAuth2/Storage/Redis.php
```diff
@@ -1,317 +1,77 @@
...
+            throw new \InvalidArgumentException('First argument to ' . __CLASS__ . ' must be an instance of Predis\Client or a configuration array');
+        }
+        
+        $this->config = array_merge($this->config, array(
+            'expire' => 0
         ), $config);
     }
 
-    protected function getValue($key)
+    protected function _makeKey($table, $key)
```
svycka: Is there a reason for underscore before method name?

afilippov1985: I removed underscores to conform PSR-2 style

afilippov1985:
@bluebaroncanada 
I think this code is ready to be merged.
Current installations that uses Cassandra, Couchbase or Redis storage will not notice changes (I hope).

afilippov1985:
@bluebaroncanada @bshaffer 
Then I want to discuss #675 (first comment) to make interfaces consistent in v2.x

PR #701, opened 2016-01-28
Fix unit tests

afilippov1985:


afilippov1985:
Travis is broken again

afilippov1985:
[Here, in comparison](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/GrantType/JwtBearer.php#L131) $jwt['exp'] casts to int, аnd overflow occurs.
```php
$d = 99999999900;
$t = time();
$b = ($d <= $t);
$i = (int)$d;

var_dump($d, $t, $b, $i);
//double(99999999900) int(1453980111) bool(false) int(1215752092) 
```

afilippov1985:
May be we should check that `access_token` key exists in array.

afilippov1985:
ping @bshaffer

bshaffer:
We should make this PR to `master` as the `develop` branch is no longer being used.

Also, the tests are passing as-is, so I'm not entirely sure what this fixes.

afilippov1985:
I'll make a few smaller PRs.

PR #709, opened 2016-03-12
Change PDO.php

vHeemstra:
When getting the default scope from the database using PDO, the function neglects the $client_id provided instead of returning the client's scope from the client table. So I added that using the getClientScope function.
So now it return either default scopes set to client from client table or returns global default scopes from scope table.

I'm not sure if this should also be changed in other storage classes.

### DIFF
--- a/src/OAuth2/Storage/Pdo.php
+++ b/src/OAuth2/Storage/Pdo.php
@@ -363,18 +363,31 @@ class Pdo implements
 
     public function getDefaultScope($client_id = null)
     {
-        $stmt = $this->db->prepare(sprintf('SELECT scope FROM %s WHERE is_default=:is_default', $this->config['scope_table']));
-        $stmt->execute(array('is_default' => true));
+        $getGlobalDefaults = true;
+        $defaultScopes = array();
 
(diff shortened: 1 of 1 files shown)


src/OAuth2/Storage/Pdo.php
```diff
@@ -363,18 +363,31 @@ public function scopeExists($scope)
...
-        $stmt->execute(array('is_default' => true));
+        $getGlobalDefaults = true;
+        $defaultScopes = array();
+
+        if (!is_null($client_id)) {
+            // Get (default) scopes from client table
+            $clientScope = $this->getClientScope($client_id);
+            if (!is_null($clientScope)) {
+                $getGlobalDefaults = false;
+                $defaultScopes = explode(' ', trim($clientScope));
```
bshaffer: let's return `$clientScopes` here and remove the `$getGlobalDefaults` variable all together. This will greatly simplify this function.

bshaffer:
We also need test coverage for this. Are you willing to add this? It will also verify if this issue is happening in the other storage classes.

PR #714, opened 2016-04-16
Phalcon Storage from the Phalcon REST API

lucasantarella:
Hi, I’ve created an integration between your library and a popular REST framework called Phalcon. I personally use the code that I’ve created in my own API, so I can attest to the fact that it works. I’m doing this to share with everyone else who uses Phalcon. Please consider adding this to your library. I know this would have saved me a huge headache when implementing this into my API.


src/OAuth2/Storage/Phalcon.php
```diff
@@ -0,0 +1,669 @@
+<?php
```
duythien: @lucasantarella I think you should created a new directory phalcon and move your code into it

lucasantarella: @duythien Where do you think I should put it? Root of the library or in the storage folder?

duythien: I have no idea, maybe in storage

PR #715, opened 2016-04-20
Oracle Storage type

WagnerSilveira:


bshaffer:
Wow this is impressive! Can we get some test coverage? I believe we can run tests locally at least, with [Oracle XE](http://www.oracle.com/technetwork/database/database-technologies/express-edition/downloads/index.html)

arthurlima96:
@bshaffer when pull request can be merged with the master branch ?

githubeing:
did you try to use the regular pdo storage with oracle? i've managed to do this, it seems to me that there's no need to have a special pdo exactly for oracle. here is how i've done that: https://github.com/bshaffer/oauth2-server-php-docs/pull/126/commits/845cca26d79ae2fd58486d6590fb2a55eaf17072

PR #723, opened 2016-05-16
Token grant issue

bkcummins:
In the event a client has no scope yet a valid scope is requested I found the access_token is granted.  Added an `} else {` to catch this.

https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Controller/TokenController.php#L175-L186


bshaffer:
@bkcummins unfortunately, the behavior of this library is that if there are no scopes returned from `$client->getClientScope`, then there are no scopes restricted. You could have this library return a fake scope (say, the string `::none::`) in order to achieve this behavior.

bkcummins:
Will do...thx!  bkc

PR #735, opened 2016-06-08
Memory Storage Fix - Add missing array_merge() parameter

steve-dave:
Hi @bshaffer,

Just a one-liner fix for a bug I stumbled on today.  I checked the other storages and Memory was the only one with the bug.

Cheers.


steve-dave:
ping @bshaffer

bshaffer:
LGTM, thanks for finding this!

PR #737, opened 2016-06-12
Update CodeIdToken.php   fix id_token for support Claims

chakphanu:
If response_type ==  "code id_token"   fix id_token for support Claims.


bshaffer:
Can you explain exactly what this PR is fixing, and ideally add a test to cover the case to confirm the issue is fixed?

bshaffer:
I added the tests for this to your commit in #749

PR #743, opened 2016-06-29
Phalcon storage

lucasantarella:


bshaffer:
Hey this is really cool! Nice work. Looks like it's essentially an ORM for the models this library uses. A few things:

- [ ] tests
- [ ] code standards (should be consistent with the rest of this library)

Does this only work for APIs or can it be used for local PDO as well?
There is a lot of boilerplate/crud code that doesn't really belong in this repo.  Is there any way to get around that, or is it required for the library to function?

lucasantarella:
Hey,
Thanks for taking a look. I'll get working on the tests, that shouldn't be to big of a problem.

I tried to adhere as closely as I could to the PDO Storage code. Could point out some places were I should make the code more consistent with the rest of the library?

By boilerplate code, do you mean the model files? Those are needed for the library to function. I don't think there is anyway to get around having those model files. I can tone down the PHPDoc if that helps at all.

src/OAuth2/Storage/Phalcon/Models/OauthClients.php
```diff
@@ -0,0 +1,92 @@
...
+    public static function findFirst($parameters = null)
+    {
+        return parent::findFirst($parameters);
+    }
+
+    /**
+     * @param mixed $parameters
+     * @return OauthUsers
+     */
+    public function getUser($parameters = null){
```
bshaffer: brace should be on a new line

src/OAuth2/Storage/Phalcon/Phalcon.php
```diff
@@ -0,0 +1,686 @@
...
+            $grant_types = explode(' ', $details['grant_types']);
+            return in_array($grant_type, (array)$grant_types);
+        }
+
+        // if grant_types are not defined, then none are restricted
+        return true;
+    }
+
+    /* OAuth2\Storage\AccessTokenInterface */
+
```
bshaffer: remove newline

src/OAuth2/Storage/Phalcon/Phalcon.php
```diff
@@ -0,0 +1,686 @@
...
+              `public_key` varchar(2000) DEFAULT NULL,
+              `private_key` varchar(2000) DEFAULT NULL,
+              `encryption_algorithm` varchar(100) DEFAULT 'RS256'
+            );
+        ";
+
+        return $sql;
+    }
+
+
```
bshaffer: remove newlines

src/OAuth2/Storage/Phalcon/Models/OauthUsers.php
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * @param mixed $parameters
+     * @return \OAuth2\Storage\Models\OauthRefreshTokens[]
+     */
+    public function getRefreshTokens($parameters = null)
+    {
+        return $this->getRelated('RefreshTokens', $parameters);
+    }
+}
```
bshaffer: add newline :P

lucasantarella: Ah, got it. Thanks for pointing them out. I'll clean it up ASAP. What do you
think about all the model files though? Too much for the style of the repo?  
  
Sent from [Nylas N1](https://link.nylas.com/link/1bn4fr7kwwyc20cix4nnlmvaw
/local-818fd84c-
52e8/0?redirect=https%3A%2F%2Fnylas.com%2Fn1%3Fref%3Dn1&r=cmVwbHkrMDAxYjM3OWI2YjFiNTYzOTQ0ZjljNDAyZGMyZTg3OTZkMWNmNzRmNjEzNDY5N2I4OTJjZjAwMDAwMDAxMTM4YzA3NWY5MmExNzBjZTA0MWQ2YWMzQHJlcGx5LmdpdGh1Yi5jb20=),
the extensible, open source mail client.  


On Jun 29 2016, at 6:02 pm, Brent Shaffer &lt;notifications@github.com&gt;
wrote:  

> In
[src/OAuth2/Storage/Phalcon/Models/OauthUsers.php](https://github.com/bshaffer/oauth2
-server-
php/pull/743#discussion_r69036739&r=cmVwbHkrMDAxYjM3OWI2YjFiNTYzOTQ0ZjljNDAyZGMyZTg3OTZkMWNmNzRmNjEzNDY5N2I4OTJjZjAwMDAwMDAxMTM4YzA3NWY5MmExNzBjZTA0MWQ2YWMzQHJlcGx5LmdpdGh1Yi5jb20=):

>  
>  
>     > +     * @return \OAuth2\Storage\Models\OauthAccessTokens[]

>     > +     */

>     > +    public function getAccessTokens($parameters = null)

>     > +    {

>     > +        return $this->getRelated('AccessTokens', $parameters);

>     > +    }

>     > +

>     > +    /**

>     > +     * @param mixed $parameters

>     > +     * @return \OAuth2\Storage\Models\OauthRefreshTokens[]

>     > +     */

>     > +    public function getRefreshTokens($parameters = null)

>     > +    {

>     > +        return $this->getRelated('RefreshTokens', $parameters);

>     > +    }

>     > +}

>  
>

> add newline :P

>

> —  
> You are receiving this because you authored the thread.  
> Reply to this email directly, [view it on
GitHub](https://github.com/bshaffer/oauth2-server-
php/pull/743/files/eedce8fc62e86eb3622d1e4b788470d7319d4128#r69036739&r=cmVwbHkrMDAxYjM3OWI2YjFiNTYzOTQ0ZjljNDAyZGMyZTg3OTZkMWNmNzRmNjEzNDY5N2I4OTJjZjAwMDAwMDAxMTM4YzA3NWY5MmExNzBjZTA0MWQ2YWMzQHJlcGx5LmdpdGh1Yi5jb20=),
or [mute the
thread](https://github.com/notifications/unsubscribe/ABs3m1yn9fmL2U6MObO_J8nn4DMZdWETks5qQutfgaJpZM4JBXTs&r=cmVwbHkrMDAxYjM3OWI2YjFiNTYzOTQ0ZjljNDAyZGMyZTg3OTZkMWNmNzRmNjEzNDY5N2I4OTJjZjAwMDAwMDAxMTM4YzA3NWY5MmExNzBjZTA0MWQ2YWMzQHJlcGx5LmdpdGh1Yi5jb20=).

bshaffer:
Thanks again for your work on this!

It would definitely be preferable to only have the single `Phalcon` storage class, without generating all the dependent wrapper classes for each table. This would keep the API consistent (i.e. passing arrays of data rather than objects) between the storage types.

Also, add the Phalcon storage class to [BaseTest](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/BaseTest.php#L7) and it'll receive all the same testing as the rest of the storage classes.

lucasantarella:
Even with PHQL, there still needs to be a model class defined for each table, even though it would be significantly cut down (only the public vars would need to be defined and the `initialize()` function). I could put all these classes at the bottom of the `Phalcon.php` file and have all the classes in one file, maybe that would help with the consistency of the repo. Let me know what you think.

lucasantarella:
I believe the test is going to fail straight-up right now. I have to add the TravisCI config for Phalcon since it's a PHP extension. Also, it's not compatible with PHP7.

bshaffer:
No need to add them all to the same file! If we need them then we'll keep them. You can skip the tests in PHP7 and add the extension in `.travis.yml`.

lucasantarella:
@bshaffer Hope you enjoyed the holiday! I'm working with one of the Phalcon dev's to get PHP 5.3 build working in Travis CI. Seems to be a travis problem about not installing the Phalcon extension. Otherwise, the tests should work fine. The library will work on any Phalcon 2.x.x and up installation, and that includes PHP >= 5.3

.travis.yml
```diff
@@ -20,11 +19,16 @@ services:
...
 - cassandra
+- mysql
+- memcached
 before_install:
-- phpenv config-rm xdebug.ini || return 0
+  - composer install --prefer-source --no-interaction
+  - vendor/bin/install-phalcon.sh 2.0.x
+  - phpenv config-rm xdebug.ini || return 0
 install:
 - composer install --no-interaction
```
bshaffer: why is `composer install` also in `install`? How about you move the phalcon installation to `install` instead?

.travis.yml
```diff
@@ -4,13 +4,12 @@ cache:
   directories:
     - $HOME/.composer/cache
     - vendor
+    - $HOME/cphalcon
 php:
 - 5.3
 - 5.4
 - 5.5
 - 5.6
-- 7
```
bshaffer: Don't remove php 7 and hhvm!! Just add an if statement around your before_install:

```yaml
  - composer install --prefer-source --no-interaction
  - if [[ ${TRAVIS_PHP_VERSION:0:1} == "5" ]]; then vendor/bin/install-phalcon.sh 2.0.x; fi;
  - phpenv config-rm xdebug.ini || return 0
```

.. Or something similar

src/OAuth2/Storage/Phalcon/Phalcon.php
```diff
@@ -664,4 +685,204 @@ public function getBuildSql($dbName = 'oauth2_server_php')
...
+/**
+ * Class PhalconConf
+ * @package OAuth2\Storage\Phalcon
+ *
+ * This config class is used purely internally for communication
+ * between the model files and the main Phalcon storage class
+ * for setting table names dynamically.
+ */
+class PhalconConf
```
bshaffer: Seems like this file could be replaced with an array and `array_merge`

lucasantarella: @bshaffer I was trying to do that, but I ran into a problem. It's really due to the nature of Phalcon.

Phalcon projects use a Dependency Injector to manage data interactions between model classes and everything else. The reason for this is that when calling model classes, its is done statically. That means there is no way to pass table names in a one-stop-shop fashion to the models unless you do it every time you statically call the `Model::find($tableName, $parameters)` or something like that. The way I accomplished this is to take a Dependency Injector `$di` and add a custom service to it and pass it on to all the models once. This way, all the models can just reference that service in the `$di` when they need to lookup table maps. The catch, though, is that a `$di` service must be an instance of a class, ANY class, so I just made a holder class to hold the information while passing to the models. 

It's not the most elegant way, but if you can find a better solution I'd seriously love to use it because I know this is the ugliest way to accomplish this task.

.travis.yml
lucasantarella: My bad... I'm kinda new to Travis and I've really only used a couple times, let alone with Phalcon. I'll fix it up right away.

src/OAuth2/Storage/Phalcon/Phalcon.php
bshaffer: Can you use `stdClass`?

lucasantarella: I believe so. I'll branch and try it out. Would this be preferable? I always liked explicitly constructed models. I'll work it out and see if it works.

lucasantarella:
Stumbling through the documentation, I found that you can use a function instead of a class. This allows the `return array()` to be used and negates the need for a special class. Sorry about the confusion! I didn't know myself.

lucasantarella:
Fixed all tests.

lucasantarella:
Hi @bshaffer I finally finished the tests and logic for this PR. What do you think about merging?

PR #748, opened 2016-07-15
added MongoDB (new php driver mongodb) storage

AstRonin:


copynpaste:
Ah I've found that #697 adds this storage too

AstRonin:
Shaffer has a choice :)

test/OAuth2/Storage/MongoDBTest.php
```diff
@@ -0,0 +1,120 @@
...
+ * @author Roman Shuplov  <astronin@gmail.com>
+ */
+class MongoDBTest extends BaseTest 
+{
+    
+    public function __construct() {
+        $mongodb = Bootstrap::getInstance()->getMongoDB();
+    }
+    
+    public function testSetClientDetails() {
```
bshaffer: curly braces should be on their own line, to be consistent with the styles in the rest of this library

test/OAuth2/Storage/MongoDBTest.php
```diff
@@ -0,0 +1,120 @@
...
+        $db->setAuthorizationCode('testcode', 'Some Client!!!', 'oauth_test_user_id', 'http://example.com', 0);
+        $this->assertEquals($db->getAuthorizationCode('testcode')['client_id'], 'Some Client!!!');
+        $db->expireAuthorizationCode('testcode');
+        $this->assertFalse($db->getAuthorizationCode('testcode'));
+        $db->setAuthorizationCode('testcode', 'Some Client', 'oauth_test_user_id', 'http://example.com', 23442);
+        $this->assertNotFalse($db->getAuthorizationCode('testcode'));
+    }
+    
+    public function testRefreshToken() {
+        $db = new MongoDB('mongodb://localhost:27017/oauth2_server_php');
```
bshaffer: The `MongoDB` object should be passed into the constructor of the storage class.

test/lib/OAuth2/Storage/BaseTest.php
```diff
@@ -1,34 +1,36 @@
...
+            array($postgres),
+            array($mongo),
+            array($mongodb),
+            array($redis),
+            array($cassandra),
+            array($dynamodb),
+            array($couchbase),
+        );
+    }
+}
```
bshaffer: hmm, something is wrong here. You must be using Windows newlines. You will need to use Unix newlines instead.

test/OAuth2/Storage/MongoDBTest.php
```diff
@@ -0,0 +1,120 @@
+<?php
+
+namespace OAuth2\Storage;
+
+/**
+ * Description of MongoDBTest
+ *
+ * @author Roman Shuplov  <astronin@gmail.com>
+ */
+class MongoDBTest extends BaseTest
```
bshaffer: The tests for the Storage Interface methods are [here](https://github.com/bshaffer/oauth2-server-php/tree/develop/test/OAuth2/Storage), so I don't think any of these tests are needed.

test/lib/OAuth2/Storage/Bootstrap.php
```diff
@@ -1,888 +1,996 @@
...
+            'client_id' => 'oauth_test_client',
+            'key'       => $this->getTestPublicKey(),
+            'subject'   => 'test_subject',
+        ));
+    }
+    
+    private function createMongoDB()
+    {
+        $bulk = new \MongoDB\Driver\BulkWrite();
+        $bulk->insert([
```
bshaffer: we need to use `array()` syntax here, since this library is compatible with PHP 5.3

bshaffer:
This looks promising, but a few major things need to be fixed before we merge this. 

1. Travis tests are failing because we need to [add the `mongodb.so` extension in travis.yml](https://github.com/GoogleCloudPlatform/getting-started-php/blob/master/.travis.yml#L42).
2. You need to fix the line endings so the diffs accurately reflect the changes submitted by this PR. check out `dos2unix` or [this SO article](http://stackoverflow.com/questions/16768776/convert-line-endings) on how to remove the carriage returns. 
3. What is the purpose of `MongoDBTest`? All these methods are covered by the standard storage suite of tests.

AstRonin:
Ok, I will do the changes after my holidays.

1. I did not have to deal with `travis.yml` ... what I should do?
2. Yes, about newlines, I noticed after pushing...
3. I needed to make sure which methodes work (write/read/remove) as I expected in different situations. For example, method `\OAuth2\Storage\Mongo::setRefreshToken` does `insert` instead of `update` and it looks like a bug. Standart tests don't identify it.

Another thing is that I can change `[]` to `array()` in `Bootstrap.php` for maintain a common style but we [cannot use new MongoDB driver](https://docs.mongodb.com/ecosystem/drivers/php/#language-compatibility) with php v5.3

bshaffer:
@AstRonin Sounds good! Okay, to address your questions:

1. To deal with `travis.yml`, add the following lines to `before_install`:

    ```yaml
    before_install:
    - if [[ ${TRAVIS_PHP_VERSION:0:3} == "7.0" ]]; then
        phpenv config-add testing/mongodb.ini;
      elif [[ ${TRAVIS_PHP_VERSION:0:3} != "5.3" ]]; then
        pecl install mongodb || /bin/true;
      fi
    ```
2. great!
3. Ok I get it. If there's a way to test all storage classes for this, that would be best. If it's a mongodb-specific thing to test, then this works great. We should remove duplicate tests though.
4. This is no problem! Have `Bootstrap::getMongoDB` use [`NullStorage`](https://github.com/bshaffer/oauth2-server-php/blob/develop/test/lib/OAuth2/Storage/Bootstrap.php#L153) in this case, and the tests will be skipped:

    ```php
    if ('5.3' === substr(phpversion(), 0, 3)) {
        $this->mongodb = new NullStorage('MongoDB', 'The mongodb.so extension is not compatible with PHP 5.3');
    }
    ```

AstRonin:
I added new commit, but travis failed again...

test/OAuth2/Storage/MongoDBTest.php
```diff
@@ -0,0 +1,132 @@
...
+        $db->unsetAccessToken('testtoken');
+        $this->assertFalse($db->getAccessToken('testtoken'));
+        $db->setAccessToken('testtoken', 'Some Client', 'oauth_test_user_id', 2);
+        $this->assertNotFalse($db->getAccessToken('testtoken'));
+        
+    }
+    
+    public function testAuthorizationCode()
+    {
+        $db = new MongoDB('mongodb://localhost:27017/oauth2_server_php');
```
robotrobot: Test fails because `Call to undefined method MongoDB\Driver\Manager::setAuthorizationCode() in /home/travis/build/bshaffer/oauth2-server-php/test/OAuth2/OpenID/Storage/AuthorizationCodeTest.php on line 77`. I think you need to be more specific here and use the namespace new \OAuth2\Storage\ MongoDB because the error suggests you are trying to call the `setAuthorizationCode` on the underlying MongoDB driver class, which obv won't work.

AstRonin:
ok, it's my fault, I mixed classes in Bootstrap in previous test...
Now error only for :
 - PHP 5.3 - MongoDB does not work in 5.3 at all, but Trevis parsed anyway...
 - HHVM - Has no extension MongoDB

I need added some thing else in travis.yml?

svycka:
php5.3 really? http://php.net/supported-versions.php

AstRonin:
https://docs.mongodb.com/ecosystem/drivers/php/#language-compatibility
It's discussed early

chadicus:
If anyone is interested, I created companion library over the weekend. https://github.com/chadicus/oauth2-storage-mongodb. 
I also needed to use MongoDB for storage, but the current restrictions of this library, it was easier to create a separate storage implementation rather than try to get a PR approved here.  As soon as this library ups its version of PHP, I'll be creating a PR.

Please keep in mind my mongodb library is in its very early stages and I'm sure it has bugs. I would not recommend for production use at this time.

bshaffer:
Added in #790

PR #762, opened 2016-09-08
Fix doc comments and double quoted strings

tmotyl:
Fixed some doc comments to make working with the project easier
in the IDE.
Also changed some strings to single quoted to prevent interpreting
backslash as escape character.


bshaffer:
awesome thank you!

PR #767, opened 2016-10-04
Error when there is no token in the request

erickfrancis:
Controlling the error but was not showing the return.


bshaffer:
This is already handled in the [`OAuth2\TokenType\Bearer`](https://github.com/bshaffer/oauth2-server-php/blob/master/src/OAuth2/TokenType/Bearer.php#L86) class, and it's correct according to the RFC.

PR #769, opened 2016-10-12
Update to current Mongo PHP Library

memmaker:
We also added the missing implementations for the JTI part.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "bshaffer/oauth2-server-php",
+    "name": "memmaker/oauth2-server-php",
     "description":"OAuth2 Server for PHP",
     "keywords":["oauth","oauth2","auth"],
     "type":"library",
@@ -9,6 +9,11 @@
             "name":"Brent Shaffer",
             "email": "bshafs@gmail.com",
             "homepage":"http://brentertainment.com"
+        },
+        {
+            "name":"lrx GmbH",
+            "email": "kontakt@lrx-solutions.de",
+            "homepage":"http://lrx-solutions,de"
         }
     ],
     "homepage": "http://github.com/bshaffer/oauth2-server-php",
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         {
             "name":"lrx GmbH",
             "email": "kontakt@lrx-solutions.de",
-            "homepage":"http://lrx-solutions,de"
+            "homepage":"http://lrx-solutions.de"
         }
     ],
     "homepage": "http://github.com/bshaffer/oauth2-server-php",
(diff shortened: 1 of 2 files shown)


chadicus:
Since the minimum PHP requirement for this library is `>=5.3.9` and the minimum requirement for the `mongodb` library is `5.4`, @bshaffer may not be too quick to merge this PR. That being said, I am working on a _component_ of sorts for implementing MongoDB storage.  You can find the library [here](https://github.com/chadicus/oauth2-storage-mongodb).  When the minimum PHP requirement of this library in increased, I will put a PR here to add the new storage.

memmaker:
This is a bit irritating. Did the Travis build fail for my PR or did it not?

chadicus:
@memmaker your build is failing because the `homepage` for your author's entry is incorrect. You have `http://lrx-solutions,de`  Note the comma before the `de`

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "bshaffer/oauth2-server-php",
+    "name": "memmaker/oauth2-server-php",
```
bshaffer: Please revert the changes made to `composer.json`

src/OAuth2/Storage/Mongo.php
```diff
@@ -17,43 +19,47 @@
...
+            'scope_table'  => 'oauth_scopes',
+            'key_table'  => 'oauth_keys',
         ), $config);
     }
 
     // Helper function to access a MongoDB collection by `type`:
     protected function collection($name)
     {
         return $this->db->{$this->config[$name]};
+        //return $this->db->{$this->config[$name]};
```
bshaffer: remove comment here

src/OAuth2/Storage/Mongo.php
```diff
@@ -17,43 +19,47 @@
...
+    PublicKeyInterface,
     OpenIDAuthorizationCodeInterface
 {
     protected $db;
     protected $config;
 
     public function __construct($connection, $config = array())
     {
-        if ($connection instanceof \MongoDB) {
+        if ($connection instanceof MongoDB\Database) {
```
bshaffer: Let's go ahead and add this to a `use` statement above.

bshaffer:
These changes use the `mongodb.so` extension instead of the `mongo.so` extension, so this can be separated in to `OAuth2/Storage/MongoDB` for BC and clarity!

chadicus:
@bshaffer  you code avoid BC changes by componentizing some of the library such as the mongodb adapter. Users could pick and choose what adapter they wanted. Also it would be easier to test the smaller libraries.

chadicus:
@memmaker I created [this](https://github.com/chadicus/oauth2-storage-mongodb) library as a temporary solution until the new MongoDB code was added to the main oauth2 server code. Feel free to use any of it's code.  I'll be abandoning it once the new MongoDB storage is merged here.

bshaffer:
I added this for BC and also fixed the tests: #790 PTAL!

I agree separate libraries would be better! I will definitely consider splitting these out.

PR #783, opened 2016-12-01
Add .gitattributes files

chadicus:
The .gitattributes file will prevent items such as test files, yaml files, etc from being included when this library is imported via composer


bshaffer:
Great! Thank you, @chadicus !

PR #784, opened 2016-12-08
RFC6750 compatibility

pgrzelka:
Changing from 'expired_token' to 'invalid_token' to be compatible with http://tools.ietf.org/html/rfc6750#section-6.2.2


bshaffer:
Thank you!

Issue #793, opened 2017-01-06
No easy way to customize JWT payload


svycka:
it is hardcoded here: https://github.com/bshaffer/oauth2-server-php/blob/v1.8.0/src/OAuth2/ResponseType/JwtAccessToken.php#L65-L75
or maybe I am missing something?

bkcummins:
I am customizing the JWT payload by extending the `OAuth2\ResponseType\JwtAccessToken` and `OAuth2\Controller\TokenController` classes which are referenced where you create your OAuth2Server.

My JwtAccessToken extension only uses the createAccessToken method, and this is where you can customize the payload.  

My TokenController extension only uses the grantAccessToken method.

Hope that helps.

svycka:
thanks @bkcummins I understand that it should be possible but would like to customize a bit easier than extending some classes

bshaffer:
@svycka I agree that we need to make this a little easier. Do you have any suggestions? One way would be to provide a callable as a config argument:

```php
$config = [
    'jwt_payload_callable' => function ($client_id, $user_id, $scope = null, $includeRefreshToken = true) { // do encoding },
];

$server = new OAuth2\Server($storage, $config);
```

Another option would be to create a specific protected method for this:

```php
class MyJwtAccessToken extends AccessToken
{
    protected encodeAccessToken($client_id, $user_id, $scope = null, $includeRefreshToken = true)
    {
         // do encoding
    }
}

//...
$server = new OAuth2\Server($storage, $config);
$server->setResponseType(new MyJwtAccessToken, 'access_token');
```

I prefer the second one, but either works for me. Thoughts?

svycka:
@bshaffer I haven't thought much, but your second example seems pretty good. Also no BC break so can be implemented now I think.

svycka:
resolved with https://github.com/bshaffer/oauth2-server-php/pull/795

hsiri:
@bshaffer sorry for hijacking this thread. but can you include this fix in your latest tag?

PR #795, opened 2017-01-10
added easier method to customize jwt token payload

svycka:
as suggested in https://github.com/bshaffer/oauth2-server-php/issues/793 
@bshaffer is it something like this?

 usage example:
```php

class MyJwtToken extends \OAuth2\ResponseType\JwtAccessToken
{
    protected function createPayload($client_id, $user_id, $scope = null)
    {
        $payload = parent::createPayload($client_id, $user_id, $scope);
        $payload['custom_param'] = 'my important value';

        return $payload;
    }
}


$server = new OAuth2\Server($storage, $config);

$jwtTokenResponseType = new MyJwtToken($publicKeyStorage, $accessTokenStorage, $refreshTokenStorage, $config);
$server->addResponseType($jwtTokenResponseType, 'token');
```

### DIFF
--- a/src/OAuth2/ResponseType/JwtAccessToken.php
+++ b/src/OAuth2/ResponseType/JwtAccessToken.php
@@ -60,7 +60,7 @@ class JwtAccessToken extends AccessToken
     public function createAccessToken($client_id, $user_id, $scope = null, $includeRefreshToken = true)
     {
         // payload to encrypt
-        $payload = $this->generatePayload($client_id, $user_id, $scope);
+        $payload = $this->createPayload($client_id, $user_id, $scope);
 
         /*
          * Encode the payload data into a single JWT access_token string
@@ -110,7 +110,7 @@ class JwtAccessToken extends AccessToken
         return $this->encryptionUtil->encode($token, $private_key, $algorithm);
     }
 
-    protected function generatePayload($client_id, $user_id, $scope = null)
+    protected function createPayload($client_id, $user_id, $scope = null)
     {
         // token to encrypt
         $expires = time() + $this->config['access_lifetime'];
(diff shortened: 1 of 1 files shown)


bshaffer:
This is great! I'm being picky here, but I don't think the term `generate` is accurate for this case. I'd prefer the method be `createPayload`. Can you make this change?

svycka:
@bshaffer done. I always have problems choosing names :) thought if we have `generateAccessToken()` `generateRefreshToken()` then why not `generatePayload()` :D not sure why those different. But anyway I did the change is it ok now? Also not sure but I guess tests are not required in this case?

bshaffer:
@svycka I don't think tests are necessary for such a small refactoring.  

As for the name, with the previous examples, the "generate" refers to the randomization of the token ID. In this case, the method is just constructing the payload and not really generating anything. That's my reasoning.

bshaffer:
Thank you for the PR!

svycka:
@bshaffer is it possible that this will be merged soon? or should I go with workaround for now?

ElectroLutz:
Is there an estimated date when this will be in a release? I find it very useful.

MajesticMaje:
Any word on this? Would be a big help!

bluebaroncanada:
@svycka, can you advise us why you closed this?

svycka:
because it is merged an will be released in new version.

bluebaroncanada:
Yeah.  Sorry.  I thought you were saying, "Closed because it's not easy to do."

bluebaroncanada:
@MajesticMaje so if you use the develop branch, you should be all set.

bshaffer:
@MajesticMaje @svycka @bluebaroncanada Sorry for the delay here, we're all tagged up! Please update your dependencies for `^1.10`:

```
composer require bshaffer/oauth2-server-php:^1.10
```

Issue #750, opened 2016-07-20
Custom JWT payload


kpn13:
Hi,

Thank you for this lib, it has save me a lot of time by now !

I'm facing a problem to custom the payload of a JWT. I think I have to override \OAuth2\ResponseType\JwtAccessToken, but I didn't find a way.

I gave my own implementation of JwtAccessToken in my Server.php in responseType field but it didn't work.

I think the problem could be createDefaultJwtAccessTokenResponseType() in OAuth2\Server.php return a default instance of JwtAccessToken without check if there is a custom JwtAccessToken in the responseTypes array. Are you agree with thas or am I totally wrong ?

Could you help me with that please ?

Thank you in advance.

afilippov1985:
Override `OAuth2\ResponseType\JwtAccessToken::createAccessToken()` or create your own class from scratch (it must implement `OAuth2\ResponseType\AccessTokenInterface`)
Then initialize `OAuth2\Server` passing 4th parameter
```php
$responseTypes = array('token' => new MyCustomJwtAccessToken(............));
$server = new OAuth2\Server($storage, $config, $grantTypes, $responseTypes);
```

achimha:
I believe it is a very common use case to add additional fields to a JWT. The other libraries I have dealt with provide simple ways to add user claims.

While I can override the `JwtAccessToken` class, I have to provide a complete implementation, copying the existing logic because there is no way to add extra fields.

galves:
You should be at least able to set the claims outlined here https://www.iana.org/assignments/jwt/jwt.xhtml

svycka:
will be easier when this will be released https://github.com/bshaffer/oauth2-server-php/pull/795

bshaffer:
This is available now in `v1.10`. See [the CHANGELOG](https://github.com/bshaffer/oauth2-server-php/blob/master/CHANGELOG.md) for details!

PR #804, opened 2017-01-27
Callable for customizing JWT payload

afilippov1985:
Even easier way for customizing JWT payload

Issue #750 
Issue #793
PR #795 

Usage:
```php
$myapp = .... // $myapp may be instance of your application or Service Container
$server = new OAuth2\Server($storage, array(
    'use_jwt_access_tokens' => true,
    'jwt_extra_payload' => function($client_id, $user_id, $scope) use ($myapp) {
        $data = $myapp->getSomeData();
        return array(
            'custom_param' => $data['something_important']
        );
    }
));
```

one more example:
```php
$myapp = .... // $myapp may be instance of your application or Service Container
$server = new OAuth2\Server($storage, array(
    'use_jwt_access_tokens' => true,
    'jwt_extra_payload' => array($myapp, 'getSomeData')
));
```

### DIFF
--- a/src/OAuth2/ResponseType/JwtAccessToken.php
+++ b/src/OAuth2/ResponseType/JwtAccessToken.php
@@ -128,15 +128,15 @@ class JwtAccessToken extends AccessToken
             'scope'      => $scope
         );
         
-        if (isset($this->config['jwt_extra_payload'])) {
-            if (!is_callable($this->config['jwt_extra_payload'])) {
-                throw new \InvalidArgumentException("config['jwt_extra_payload'] is not callable");
+        if (isset($this->config['jwt_extra_payload_callable'])) {
+            if (!is_callable($this->config['jwt_extra_payload_callable'])) {
+                throw new \InvalidArgumentException("config['jwt_extra_payload_callable'] is not callable");
             }
             
-            $extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);
+            $extra = call_user_func($this->config['jwt_extra_payload_callable'], $client_id, $user_id, $scope);
             
             if (!is_array($extra)) {
-                throw new \InvalidArgumentException("config['jwt_extra_payload'] callable must return array");
+                throw new \InvalidArgumentException("config['jwt_extra_payload_callable'] callable must return array");
             }
             
             $payload = array_merge($extra, $payload);
--- a/src/OAuth2/ResponseType/JwtAccessToken.php
+++ b/src/OAuth2/ResponseType/JwtAccessToken.php
@@ -130,13 +130,13 @@ class JwtAccessToken extends AccessToken
         
         if (isset($this->config['jwt_extra_payload_callable'])) {
             if (!is_callable($this->config['jwt_extra_payload_callable'])) {
-                throw new \InvalidArgumentException("config['jwt_extra_payload_callable'] is not callable");
+                throw new \InvalidArgumentException('jwt_extra_payload_callable is not callable');
             }
             
             $extra = call_user_func($this->config['jwt_extra_payload_callable'], $client_id, $user_id, $scope);
             
             if (!is_array($extra)) {
-                throw new \InvalidArgumentException("config['jwt_extra_payload_callable'] callable must return array");
+                throw new \InvalidArgumentException('jwt_extra_payload_callable must return array');
             }
             
             $payload = array_merge($extra, $payload);
--- a/src/OAuth2/ResponseType/JwtAccessToken.php
+++ b/src/OAuth2/ResponseType/JwtAccessToken.php
@@ -129,16 +129,17 @@ class JwtAccessToken extends AccessToken
         );
         
         if (isset($this->config['jwt_extra_payload'])) {
-            if (is_callable($this->config['jwt_extra_payload'])) {
-                $extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);
-                if (is_array($extra)) {
-                    $payload = array_merge($extra, $payload);
-                } else {
-                    throw new \InvalidArgumentException("config['jwt_extra_payload'] callable must return array");
-                }
-            } else {
+            if (!is_callable($this->config['jwt_extra_payload'])) {
                 throw new \InvalidArgumentException("config['jwt_extra_payload'] is not callable");
             }
+            
+            $extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);
+            
+            if (!is_array($extra)) {
+                throw new \InvalidArgumentException("config['jwt_extra_payload'] callable must return array");
+            }
+            
+            $payload = array_merge($extra, $payload);
         }
         
         return $payload;
--- a/src/OAuth2/ResponseType/JwtAccessToken.php
+++ b/src/OAuth2/ResponseType/JwtAccessToken.php
@@ -116,7 +116,7 @@ class JwtAccessToken extends AccessToken
         $expires = time() + $this->config['access_lifetime'];
         $id = $this->generateAccessToken();
 
-        return array(
+        $payload = array(
             'id'         => $id, // for BC (see #591)
             'jti'        => $id,
             'iss'        => $this->config['issuer'],
@@ -127,5 +127,15 @@ class JwtAccessToken extends AccessToken
             'token_type' => $this->config['token_type'],
             'scope'      => $scope
         );
+        
+        if (isset($this->config['jwt_extra_payload']) && is_callable($this->config['jwt_extra_payload'])) {
+            $func = $this->config['jwt_extra_payload'];
+            $extra = $func($client_id, $user_id, $scope);
+            if (is_array($extra)) {
+                $payload = array_merge($extra, $payload);
+            }
+        }
+        
+        return $payload;
     }
 }
--- a/src/OAuth2/ResponseType/JwtAccessToken.php
+++ b/src/OAuth2/ResponseType/JwtAccessToken.php
@@ -128,11 +128,14 @@ class JwtAccessToken extends AccessToken
             'scope'      => $scope
         );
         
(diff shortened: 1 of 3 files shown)


src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,15 @@ protected function createPayload($client_id, $user_id, $scope = null)
             'token_type' => $this->config['token_type'],
             'scope'      => $scope
         );
+        
+        if (isset($this->config['jwt_extra_payload']) && is_callable($this->config['jwt_extra_payload'])) {
+            $func = $this->config['jwt_extra_payload'];
+            $extra = $func($client_id, $user_id, $scope);
```
svycka: maybe better use `call_user_func()` instead of variable?

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,15 @@ protected function createPayload($client_id, $user_id, $scope = null)
             'token_type' => $this->config['token_type'],
             'scope'      => $scope
         );
+        
+        if (isset($this->config['jwt_extra_payload']) && is_callable($this->config['jwt_extra_payload'])) {
```
afilippov1985: I agree

src/OAuth2/ResponseType/JwtAccessToken.php
afilippov1985: Yes! My mistake.

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,15 @@ protected function createPayload($client_id, $user_id, $scope = null)
...
+            $func = $this->config['jwt_extra_payload'];
+            $extra = $func($client_id, $user_id, $scope);
+            if (is_array($extra)) {
+                $payload = array_merge($extra, $payload);
```
afilippov1985: Do you mean that callable must be defined like this? (pay attention that `$payload` passed by reference)
`function(&$payload, $client_id, $user_id, $scope) {}`

I think that standard JWT claims should be protected from external changes. That's why I put `$payload` last in `array_merge()`

svycka: no, I man instead of:
```php
$extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);
if (is_array($extra)) {
    $payload = array_merge($extra, $payload);
}
```
do this:
```php
$payload = call_user_func($this->config['jwt_extra_payload'], $payload);
```

svycka: I don't see why can't change those if we need. specification does not say they are required and even if this lib does require some when we override them here it is most likely we will handle this change in other places as well.
@bshaffer what do you think?

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,18 @@ protected function createPayload($client_id, $user_id, $scope = null)
...
         );
+        
+        if (isset($this->config['jwt_extra_payload'])) {
+            if (is_callable($this->config['jwt_extra_payload'])) {
+                $extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);
+                if (is_array($extra)) {
+                    $payload = array_merge($extra, $payload);
+                }
+            } else {
+                throw new \InvalidArgumentException("config['jwt_extra_payload'] is not callable");
```
svycka: because of `if (isset($this->config['jwt_extra_payload'])) ` will throw exception on empty values like `null, '', 0 or false`

svycka:
I think this is new feature and requires tests.

src/OAuth2/ResponseType/JwtAccessToken.php
afilippov1985: Doing so `callable` may break JWT. Standard JWT claims should be protected from external changes.

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,20 @@ protected function createPayload($client_id, $user_id, $scope = null)
             'token_type' => $this->config['token_type'],
             'scope'      => $scope
         );
+        
+        if (isset($this->config['jwt_extra_payload'])) {
+            if (is_callable($this->config['jwt_extra_payload'])) {
+                $extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);
+                if (is_array($extra)) {
```
svycka: I don't think this is required

src/OAuth2/ResponseType/JwtAccessToken.php
svycka: this feature is to generate those claims for example if you would want increase jwt token expire time for some user you could not. But if you only want to add new parameters then yes its ok

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,20 @@ protected function createPayload($client_id, $user_id, $scope = null)
             'token_type' => $this->config['token_type'],
             'scope'      => $scope
         );
+        
+        if (isset($this->config['jwt_extra_payload'])) {
+            if (is_callable($this->config['jwt_extra_payload'])) {
```
bshaffer: I would prefer these exceptions be handled like this:

```php
if (!is_callable($this->config['jwt_extra_payload']))) {
    throw new \InvalidArgumentException("config['jwt_extra_payload'] is not callable");
}

$extra = call_user_func($this->config['jwt_extra_payload'], $client_id, $user_id, $scope);

if (!is_array($extra)) {
    throw new \InvalidArgumentException("config['jwt_extra_payload'] callable must return array");
}

$payload = array_merge($extra, $payload);
```

src/OAuth2/ResponseType/JwtAccessToken.php
bshaffer: Yes it is, otherwise the `array_merge` below will throw an error.

src/OAuth2/ResponseType/JwtAccessToken.php
svycka: @bshaffer I still don't like the idea that it is impossible to change values in $payload.

src/OAuth2/ResponseType/JwtAccessToken.php
svycka: @bshaffer this is about array_merge error vs custom error? result will be the same :D but maybe yes more clear error message is better

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -129,16 +129,17 @@ protected function createPayload($client_id, $user_id, $scope = null)
         );
         
         if (isset($this->config['jwt_extra_payload'])) {
```
svycka: since I dont like nesting I would change this to 
```php
if (isset($this->config['jwt_extra_payload'])) {
    return $payload;
}
//...
```

afilippov1985:
Let's vote:

- [x] Defensive programming
`$payload = array_merge($extra, $payload);`

OR

- [ ] Flexibility
`$payload = array_merge($payload, $extra);`

src/OAuth2/ResponseType/JwtAccessToken.php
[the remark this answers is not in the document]
bluebaroncanada: I don't like any of this.  It violates the paradigms the library was built on.  I love doing callbacks like this, but it ought to be consistent.  If it's coming down to just an array merge, I think #795 is the place to do this.

bshaffer:
This is looking great. We need to add the parameter [here](https://github.com/bshaffer/oauth2-server-php/pull/804/files#diff-e1a1ef61268d9718737b5d1a0792115bR140) also!

afilippov1985:
How about to change parameter name to `jwt_extra_payload_callable` to clarify its purpose?
May be some other name?

bshaffer:
@afilippov1985 yes, I think `jwt_extra_payload_callable` works for me.

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,21 @@ protected function createPayload($client_id, $user_id, $scope = null)
             'token_type' => $this->config['token_type'],
             'scope'      => $scope
         );
+        
+        if (isset($this->config['jwt_extra_payload_callable'])) {
+            if (!is_callable($this->config['jwt_extra_payload_callable'])) {
+                throw new \InvalidArgumentException("config['jwt_extra_payload_callable'] is not callable");
```
bshaffer: nit, but I'd rather this just be "jwt_extra_payload_callable is not callable"

src/OAuth2/ResponseType/JwtAccessToken.php
```diff
@@ -127,5 +127,21 @@ protected function createPayload($client_id, $user_id, $scope = null)
...
+            }
+            
+            $extra = call_user_func($this->config['jwt_extra_payload_callable'], $client_id, $user_id, $scope);
```
bshaffer: Why don't we just pass this function `$payload` and expect the returned array to be the new payload, e.g.:

```php
$payload = call_user_func($this->config['jwt_extra_payload_callable'], $payload);
```

afilippov1985: This may break proper work of JWT
See our conversation with @svycka - https://github.com/bshaffer/oauth2-server-php/pull/804/files/86bd64e745ee34fa03bfb3456e5d2a2ab0a1c07e#r98169436
I agree to replace this
`$payload = array_merge($extra, $payload);`
to this
`$payload = array_merge($payload, $extra);`

afilippov1985: then, we should also rename the parameter to `jwt_override_payload_callable`

bshaffer: I'm convinced. Ty for the explanation.

PR #806, opened 2017-01-30
Bearer::getAccessTokenParameter: added application/json into content-type evaluation

edegaudenzi:
It can be useful to add 'application/json' together with 'application/x-www-form-urlencoded' in the content-type white list. 
From what written in the resource _http://tools.ietf.org/html/rfc6750#section-2.2_ pointed in the code comment, the IETF clearly specifies to use 'application/x-www-form-urlencoded'

> The HTTP request entity-header includes the "Content-Type" header field set to "application/x-www-form-urlencoded".

If there are no added security issues that I cannot see at the moment, you can merge the pull request.

### DIFF
--- a/src/OAuth2/TokenType/Bearer.php
+++ b/src/OAuth2/TokenType/Bearer.php
@@ -113,7 +113,7 @@ class Bearer implements TokenTypeInterface
                 $contentType = substr($contentType, 0, $pos);
             }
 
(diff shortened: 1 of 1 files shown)


bshaffer:
Tests are failing!

bshaffer:
Currently, when pulling the access token from the POST body, the content type does not support JSON. To support JSON, send the access token in with the `Authorization` header, or in the querystring.

PR #807, opened 2017-02-02
removed unused class attribute

svycka:


src/OAuth2/OpenID/Controller/UserInfoController.php
```diff
@@ -16,8 +16,6 @@
  */
 class UserInfoController extends ResourceController implements UserInfoControllerInterface
 {
-    private $token;
-
     protected $tokenType;
```
svycka: also I don't know the reason why need here those:
```
    protected $tokenType;
    protected $tokenStorage;
    protected $config;
    protected $scopeUtil;
```

PR #819, opened 2017-04-07
CORS Following (issue while making request with axios)

Unkodero:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

### DIFF
--- a/src/OAuth2/Controller/TokenController.php
+++ b/src/OAuth2/Controller/TokenController.php
@@ -97,9 +97,15 @@ class TokenController implements TokenControllerInterface
      */
     public function grantAccessToken(RequestInterface $request, ResponseInterface $response)
     {
+        if (strtolower($request->server('REQUEST_METHOD')) == 'options') {
+            $response->addHttpHeaders(array('Allow' => 'POST, OPTIONS'));
+
+            return null;
+        }
+        
         if (strtolower($request->server('REQUEST_METHOD')) != 'post') {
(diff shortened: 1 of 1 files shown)


src/OAuth2/Controller/TokenController.php
```diff
@@ -97,9 +97,15 @@ public function handleTokenRequest(RequestInterface $request, ResponseInterface
...
     {
+        if (strtolower($request->server('REQUEST_METHOD')) == 'options') {
+            $response->addHttpHeaders(array('Allow' => 'POST, OPTIONS'));
+
+            return null;
+        }
+        
         if (strtolower($request->server('REQUEST_METHOD')) != 'post') {
             $response->setError(405, 'invalid_request', 'The request method must be POST when requesting an access token', '#section-3.2');
-            $response->addHttpHeaders(array('Allow' => 'POST'));
```
ApacheEx: what is the reason of deletion this line? it should be here. Everything else looks good.

bshaffer: Instead, this should be `array('Allow' => array('POST', 'OPTIONS'))`

bshaffer:
Can we get a Test Case added for this?

src/OAuth2/Controller/TokenController.php
ApacheEx: Yeah, that's true

ApacheEx:
@bshaffer,
I've added test cases. Please check #829

bshaffer:
This is great! #829 is merged

PR #834, opened 2017-05-27
use random_bytes if available

oittaa:


bshaffer:
is `random_bytes` the new hotness, and considered better than `openssl_random_pseudo_bytes`?

soren121:
@bshaffer Yes, for reasons listed here: [paragonie/random_compat#96](https://github.com/paragonie/random_compat/issues/96)

PR #846, opened 2017-06-16
Update error code from 400 to 401

ZhouShiJun5:
Update error code from 400 (Bad Request) to 401 (Unauthorized)


bshaffer:
`401` Is not the appropriate status code here:

```
Similar to 403 Forbidden, but specifically for use when authentication is required and has failed or has not yet been provided
```
`401` is a request to a *resource* which has failed because authentication hasn't been provided. In the case above, authentication has failed because the request being made is wrong, i.e. BAD REQUEST.

PR #851, opened 2017-07-18
Fix PHPDoc

maxailloud:
Basically just some PHPDoc fix.

### DIFF
--- a/src/OAuth2/Controller/TokenController.php
+++ b/src/OAuth2/Controller/TokenController.php
@@ -156,8 +156,8 @@ class TokenController implements TokenControllerInterface
          * ClientAssertionTypes allow for grant types which also assert the client data
          * in which case ClientAssertion is handled in the validateRequest method
          *
-         * @see OAuth2\GrantType\JWTBearer
-         * @see OAuth2\GrantType\ClientCredentials
+         * @see \OAuth2\GrantType\JWTBearer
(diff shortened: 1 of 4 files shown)


bshaffer:
Dumb question, is the leading `\` necessary? I assume the answer is PHPDoc reads from the same namespace as the executing PHP code does, and since we don't have a `use` statement and we are in a different namespace, PHPDoc sees this as `\OAuth2\Controller\OAuth2\GrantType\ClientCredentials` (for [instance](https://github.com/Chargifi/oauth2-server-php/blob/5a26917d578c99e969ed0b008106af4145586475/src/OAuth2/Controller/TokenController.php#L160))

maxailloud:
In PHPStorm not putting the leading `\` gives me an error. 
And rightfully as the class `\OAuth2\Controller\OAuth2\GrantType\ClientCredentials` doesn't exist, but `\OAuth2\GrantType\ClientCredentials` does.

bluebaroncanada:
If you use `use`, you don't need the preceding `\`.

maxailloud:
Yes but importing a class just for the PHPDoc is pointless.
It's just updating the namespace of the classes use in the PHPDoc to have the real namespace.

bluebaroncanada:
Yeah.  Sorry.  I think that made some of you think that I actually care. :p

PR #856, opened 2017-07-26
fix AuthorizeController scope error field

abemassry:
According to the rfc https://tools.ietf.org/html/rfc6749#section-4.1.2.1
scope errors should have an error field of invalid_scope this fixes that
and the test.  The TokenController has the same error description but
uses the invalid_scope error so this commit makes AuthorizeController
match that.


bshaffer:
Because this breaks backwards compatibility, this should be submitted to the v2 branch!

abemassry:
@bshaffer opened up a new PR #870 against the v2 branch, closing this one.

PR #857, opened 2017-07-27
Fix buildUri for URIs with fragments

mneumueller:
Output the query part after the fragment part, otherwise redirect URIs with fragment don't work.

Example:
Given Redirect URI: https://www.example.com/#/oauth2
Result of buildUri before change: https://www.example.com/?code=2sadkfa3378dsadj&state=1#/oauth2
Result after change: https://www.example.com/#/oauth2?code=2sadkfa3378dsadj&state=1

### DIFF
--- a/src/OAuth2/Controller/AuthorizeController.php
+++ b/src/OAuth2/Controller/AuthorizeController.php
@@ -376,8 +376,8 @@ class AuthorizeController implements AuthorizeControllerInterface
             . ((isset($parse_url["host"])) ? $parse_url["host"] : "")
             . ((isset($parse_url["port"])) ? ":" . $parse_url["port"] : "")
             . ((isset($parse_url["path"])) ? $parse_url["path"] : "")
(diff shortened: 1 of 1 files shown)


danopz:
This is application specific. The fragment MUST be the last part of the URI and the query part is before that. You try to add the query part inside of the fragment part - all after # is fragment - which results in an URI without query params.
So how are we able to get https://www.example.com/?code=2sadkfa3378dsadj&state=1#/oauth2?

Spomky:
Furthermore to the answer provided by @danopz and as per the [RFC6749 section 3.1.2](http://tools.ietf.org/html/rfc6749#section-3.1.2), the redirection URI must not contain a fragment parameter.
The given redirect URI (https://www.example.com/#/oauth2) should be rejected by the server.

mneumueller:
Thanks for the clarification and the the RFC link.
Unfortunately I receive as Redirect URI https://www.example.com/#/oauth2 from an Angular2 App.
I'll keep my fork until this is fixed in the Angular2 App.

Issue #871, opened 2017-09-17
Parameter must be an array or an object that implements Countable


TomMettam:
Using this module with PHP 7.2 triggers a warning:

'message' => 'count(): Parameter must be an array or an object that implements Countable',
  'file' => 'vendor/bshaffer/oauth2-server-php/src/OAuth2/Server.php',

with package version v1.9.0

bshaffer:
thanks for reporting the issue and providing the fix! You are what makes open source great and I salute you! 🖖

PR #872, opened 2017-09-17
Fix for https://github.com/bshaffer/oauth2-server-php/issues/871

TomMettam:
Simple fix for #871

### DIFF
--- a/src/OAuth2/Server.php
+++ b/src/OAuth2/Server.php
@@ -93,7 +93,7 @@ class Server implements ResourceControllerInterface,
     /**
      * @var array
      */
-    protected $responseTypes;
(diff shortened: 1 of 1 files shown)


bluebaroncanada:
I traced all usages of this variable and I think this is safe, though, I think what might be better is changing line 525 from `if (0 == count($this->responseTypes)) {` to `if (!isset($this->responseTypes) || 0 == count($this->responseTypes)) {`.  This will have fewer effects on the code.

bshaffer:
This looks good, but I am adding php 7.2 in https://github.com/bshaffer/oauth2-server-php/pull/873 to verify the fix

PR #873, opened 2017-09-19
adds php 7.2 to travis

bshaffer:


bluebaroncanada:
I'm worried this commit or commits like it could cause other logical statements to trigger.  For example: null or truth checks.

bluebaroncanada:
It might not just be this code that's affected by this change, too.

bluebaroncanada:
Never mind.  It shouldn't be an issue.  My apologies.

bluebaroncanada:
Err.  Actually it is if `$a=[]; $a===null;`

bshaffer:
You bring up a good point. `$server->getResponseType` and `$server->getGrantType` will be returning differently with this change. It may be time we rename `master` to `v1-master`, and rename `develop` to `master` and merge in the changes from the `v2` branch.

bshaffer:
@bluebaroncanada This change is very benign. We should look into moving to `v2.0` soon, however!

PR #899, opened 2017-12-06
[Fix test] "Expiration (exp) time must be a unix time stamp" on 32-bit platform

afilippov1985:
After this PR #898 I able to run tests. And I get:
```
There were 2 failures:

1) OAuth2\GrantType\JwtBearerTest::testInvalidJti
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Expiration (exp) time must be a unix time stamp'
+'JSON Token Identifier (jti) has already been used'

D:\eclipseworkspace\oauth2-server-php\test\OAuth2\GrantType\JwtBearerTest.php:291

2) OAuth2\GrantType\JwtBearerTest::testJtiReplayAttack
Failed asserting that null is not null.
```

Tests failed because `ctype_digit($jwt['exp'])` evals to `false` when `$jwt['exp']` is `double` (it becomes `double` when `$jwt['exp'] > PHP_INT_MAX`)
```php
// on 32-bit
$d = 99999999900;
var_dump($d, ctype_digit($d));
//double(99999999900) bool(false) 

$d = '99999999900';
var_dump($d, ctype_digit($d));
//string(11) "99999999900" bool(true) 

// on 64-bit
$d = 99999999900;
var_dump($d, ctype_digit($d));
//int(99999999900) bool(true) 

$d = '99999999900';
var_dump($d, ctype_digit($d));
//string(11) "99999999900" bool(true) 
```

To fix this we should always write big integers as strings (`'99999999900'`)
or write integers as integers (`12345`) and integers must be <= PHP_INT_MAX

### DIFF
--- a/test/OAuth2/GrantType/JwtBearerTest.php
+++ b/test/OAuth2/GrantType/JwtBearerTest.php
@@ -282,7 +282,7 @@ EOD;
         $server = $this->getTestServer();
         $request = TestRequest::createPost(array(
                 'grant_type' => 'urn:ietf:params:oauth:grant-type:jwt-bearer',  // valid grant type
-                'assertion' => $this->getJWT(99999999900, null, 'testuser@ourdomain.com', 'Test Client ID', 'used_jti'), // valid assertion with invalid scope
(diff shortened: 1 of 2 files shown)


bshaffer:
Thanks, I did not really test this library for 32 bit. My deepest regrets that you are dealing with such.

Issue #900, opened 2017-12-07
PHP 5.4 array syntax


augsteyer:
Was running the code on PHP5.3 and got an error in my logs:
```
[Thu Dec 07 05:19:10 2017] [error] [client 127.0.0.1] FastCGI: server "/usr/lib/cgi-bin/php5-fcgi" stderr: PHP message: PHP Parse error:  syntax error, unexpected '[' in /var/www/mage1810/lib/Module/cloud-integration-magento-oauth2/vendor/bshaffer/oauth2-server-php/src/OAuth2/Server.php on line 91
```
These are the php5.4 syntax culprits:
```
/**
     * @var array
     */
    protected $grantTypes = [];

    /**
     * @var array
     */
    protected $responseTypes = [];
```
Based on the required composer version of php 5.3.9, I am assuming this is a bug.

svycka:
hmm seems like @bshaffer forgot to remove php 5.3 support in https://github.com/bshaffer/oauth2-server-php/pull/869 not sure what to do now. removing support now will leave old users with a broken latest version that supports php 5.3.

but @augsteyer maybe time to update your php version? php5.3 is almost 4 years is not supported

augsteyer:
I believe this is the only thing that keeps this from being 5.3 compatible. Easier to switch these two to ```= array(); ``` and retain 5.3.9 support, no?
Also, when working with 100k+ merchant systems, it's not that easy to tell everyone to upgrade their php. Merchant's are risk aversive bunch and would rather avoid our plugin than upgrade their system.

bshaffer:
This has been merged into `master`. Thanks for the fix!

PR #901, opened 2017-12-07
#900 Removing PHP 5.4 shorthand array syntax

augsteyer:


svycka:
then maybe add back php 5.3 to travis test matrix to be sure that everything works

bshaffer:
Works for me, thank you!

gmox:
out of curiosity, why support a version of PHP that's not been updating for over 2 years?

augsteyer:
If you provide a plugin/service to a million+ online retailers out there, there is still a large number of them who have not upgraded their servers because of fears of breaking their store. One can say to not support such merchants out of principle, but that would mean losing millions of dollars. My CFO's always go for the dollars :) 

p.s. heck, we still have merchants on 5.2

Ben-Ho:
@bshaffer do you plan on creating a new release for this fix? Currently we include your package by directly referencing the commit-hash in our projects composer.json, because we thought this would just be a matter of time. (currently no big problem as there are not many new changes in rou repo, but it gives me a bad feeling ;-) )

Ben-Ho:
@bshaffer thank you! :-)

PR #907, opened 2018-01-04
try to support openid connect subjects identifier types, there have no finish test.

hhniao:
just show my purpose.
if it is worth to do, i will finish all my work.


hhniao:
it is strange!  why not pass test? 
ClientTest::testCheckRestrictedGrantType i have no anything is modified about this.

PR #911, opened 2018-01-12
[908] - remove not null check in order by

TheArtOfPour:
The *order by desc* clause will already ensure rows with client id populated are preferred.


JamesB7:
That is only true in MySQL. Many databases order nulls higher than values.

bshaffer:
closing as this changes expected behavior and I do not see the benefit.

PR #919, opened 2018-03-02
Resolve #918 "With use_jwt_access_tokens=true, token type is lowercase"

### LINKED ISSUE #918: With use_jwt_access_tokens=true, token type is lowercase
With configuration "use_jwt_access_tokens" set to "true", the resulting "token_type" is "bearer". Shouldn't it be "Bearer"?

hardysim:
closes #918


hardysim:
This is the only place where I've spotted a lower-case "bearer". Everywhere else it's already starting with a capital `B`.

It would be good if someone could verify that this really resolves #918. Thx.

bshaffer:
see my comments in #918 for why this is not needed.

Issue #523, opened 2015-02-19
PSR-7


bshaffer:
We should remove our Request/Response interfaces in favor of PSR-7:

https://github.com/php-fig/fig-standards/blob/master/proposed/http-message.md

zerrvox:
Any timeline on this?

bshaffer:
None yet, but thank you for voicing your interest

sumeko:
+1

oytuntez:
This was among my notes about this package, great it's on the roadmap.

benbor:
+1

bshaffer:
here's some [great reasons to switch](http://slides.mwop.net/2015-04-10-PSR7-in-the-Middle/#/6). I'll be working on this soon.  [Zend's Diactoros](https://github.com/zendframework/zend-diactoros) seems to be the best library for implementing the actual request object, although this eventually will matter very little, as this library fits the middleware profile perfectly.

bshaffer:
There is a question of how to implement error handling. The Response object is now immutable, which means the current model will not work. We will need to make the following changes:

1. **AuthorizeController::handleAuthorizeRequest** - returns `ResponseInterface`
1. **AuthorizeController::validateAuthorizeRequest** - returns `true` on success, `ResponseInterface` on error
1. **ResourceController::verifyResourceRequest** - returns `true` on success, `ResponseInterface` on error
1. **ResourceController::getAccessTokenData**- returns `array` on success, `ResponseInterface` on error
1. **TokenController::handleTokenRequest**- `ResponseInterface` 
1. **TokenController::grantAccessToken**- returns `array` on success, `ResponseInterface` on error
1. **TokenController::revoke**- returns `true` on success, `ResponseInterface` on error

There is a pattern forming here, where you get a response object on error. This isn't a terrible solution, but is there a better one?

copynpaste:
Maybe we should just throw some Exceptions on error and return the expected type.
As right now errors result in different responses - so it will be possible using Exceptions and transform them to that response on catch.

bshaffer:
@copynpaste I do like this solution. It's also the model implemented by [firebase/jwt](https://github.com/firebase/php-jwt). However, the general sentiment exists that `Exception` classes should not be used for error handling and validation. Also, requiring users to surround certain function calls with `try-catch` is a pain.

[...]

PR #928, opened 2018-05-16
Make the ResponseInterface Symfony 4 compliant

X-Coder264:
This is a breaking change as the interface is now type-hinted which means that the min PHP version had to be bumped (so this should be tagged as the next major version).

Fixes https://github.com/bshaffer/oauth2-server-httpfoundation-bridge/issues/31

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,6 @@ cache:
     - $HOME/.composer/cache
     - vendor
 php:
-- 5.3
-- 5.4
-- 5.5
-- 5.6
-- 7.0
 - 7.1
 - 7.2
 env:
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "psr-0": { "OAuth2": "src/" }
     },
     "require":{
-        "php":">=5.3.9"
+        "php":"^7.1.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.0",
--- a/src/OAuth2/Response.php
+++ b/src/OAuth2/Response.php
@@ -140,13 +140,13 @@ class Response implements ResponseInterface
     }
 
     /**
-     * @param int $statusCode
-     * @param string $text
+     * @param int              $statusCode
+     * @param string|bool|null $text
      * @throws InvalidArgumentException
      */
-    public function setStatusCode($statusCode, $text = null)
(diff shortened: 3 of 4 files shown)


danopz:
This would not really make it to core. The plan is already to [go with PSR-7](https://github.com/bshaffer/oauth2-server-php/issues/523), there is just no implementation right now.

X-Coder264:
@danopz The plan to go with PSR-7 is great, but it's been 3 years since any development was done in order to achieve that goal. So it's obvious that the plan is either dead or at the very least it isn't gonna be achieved anytime soon. Meanwhile this is the only blocker that we have on our projects in order to update them to use the latest version of Laravel or Symfony (and obviously we are not the only ones). So I'd say that this is worth merging and tagging as a new (major) version. Because the only other option is to basically refactor/rewrite the entire OAuth part of our projects to use another package which is a huge task (or we should forever stay on an outdated framework version just because of this which just isn't an option).

janzankowski:
+1 For @X-Coder264 argument. This should be fixed.

bshaffer:
Question - why is this a blocker for those projects? These interfaces are forwards-compatible with the latest versions of PHP.

bshaffer:
I understand now, the interfaces were updates as part of the symfony package.

We may need to do some sneaky autoloading to get around this issue while maintaining BC with previous versions of PHP. Any ideas?

X-Coder264:
IMHO I don't think there is any need to do any "sneaky" stuff just to preserve BC with ancient PHP versions. As of today even PHP 7.0 reached EOL and PHP 5.6 extended support period ends at the end of the month.

bshaffer:
@X-Coder264 unfortunately, in practice, those other versions of PHP are heavily used and I don't want to drop support for them just yet.

X-Coder264:
Well, we could still go with the route of tagging a new major version with this PR and then backporting all important bug fixes etc. to the 1.x branch. That seems like a better way to approach this problem that to be doing some hacky stuff just so that it can work with all PHP versions.

bshaffer:
Added https://github.com/bshaffer/oauth2-server-httpfoundation-bridge/pull/34, which will fix this issue without changing interfaces or requiring autoload hacks ;)

Issue #363, opened 2014-04-02
No easy way to customize JWT header


bshaffer:
The [`Jwt`](https://github.com/bshaffer/oauth2-server-php/blob/develop/src/OAuth2/Encryption/Jwt.php) encryption class has the header hardcoded in the `encode` function, and requires quite a bit to override.  

We should make `sign` protected, and make the headers editable (or pass them in when you encode the token in the first place)

svycka:
implemented with https://github.com/bshaffer/oauth2-server-php/pull/795

clubery:
@svycka @bshaffer when will #795 be released?

svycka:
it is merged so in next release

BVMiko:
@svycka I believe that you may be confusing the JWT _header_ from the JWT _payload_.  #795 made it easier to modify the payload; however the JWT header is generated later during the `encodeToken` phase.

svycka:
@BVMiko yes, you are right this is for payload, not a header

menturion:
The kid must be placed in the JWT header of the ID token.

natecovington:
I believe we have this fixed downstream in the Drupal Oauth2_Server module:

https://github.com/matrix-org/synapse/issues/13747#issuecomment-1268719703
https://www.drupal.org/project/oauth2_server/issues/2288237#comment-14737774

PR #932, opened 2018-05-22
Include Client ID (kid) parameter in the JWT header

BVMiko:
Currently, `Oauth2\Encryption\FirebaseJwt::encode` has a different signature from `OAuth2\Encryption\Jwt::encode`; the `FirebaseJwt` function has an additional `$keyId` parameter.  This parameter, when included, is passed into the `kid` parameter of the JWT header.

Unfortunately, neither of the two classes which use it (`OAuth2\OpenID\ResponseType\IdToken` and `OAuth2\ResponseType\JwtAccessToken`) include the parameter by default, so a lot of code is required to get this parameter added into the JWT header.

This PR modifies the signature of `OAuth2\Encryption\EncryptionInterface::encode` to include the `$keyId` parameter, thus standardizing between the two provided encryption implementations to offer a way to incorporate the `kid` parameter.  It also modifies the two ResponseTypes which use the `encode` method to pass the `$client_id`.

There is also one more related issue included in this PR: modern versions of firebase/php-jwt use namespaces, while the references in `Oauth2\Encryption\FirebaseJwt` do not.  I've modified this class to use the proper namespace parameters, and also upgraded the minimum suggestion of firebase/php-jwt package from ~2.2 (released 2015-06-22) to ~3.0 (released 2015-07-22).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "require-dev": {
         "phpunit/phpunit": "^4.0",
         "aws/aws-sdk-php": "~2.8",
-        "firebase/php-jwt": "~2.3",
+        "firebase/php-jwt": "~3.0",
         "predis/predis": "dev-master",
         "thobbs/phpcassa": "dev-master",
         "mongodb/mongodb": "^1.1"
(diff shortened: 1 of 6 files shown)


BVMiko:
I also want to reference to issue #363; though this PR doesn't offer a way to modify the JWT header with arbitrary data; it does automatically provide the `kid` parameter; which I think may solve the issue for most.

menturion:
+1

Issue #940, opened 2018-07-13
OAuth2 server sometimes returns "expires_in" as a string


jlfranklin:
When used as part of the Drupal [OAuth2 Server](https://www.drupal.org/project/oauth2_server) module, the token's "expires_in" value may be returned as a JSON String, not a JSON Number.  Technically, this violates the spec, and some [unforgiving client libraries](https://github.com/google/google-oauth-java-client/issues/62) will refuse to accept an improperly formatted token.

bshaffer:
As mentioned in the referenced PR, `expires_in` is set as a configuration value on the `OAuth2\Server` object, so if this is being set as a string it will be returned as a string. I imagine there's a bug in the upstream Drupal module causing this.

jlfranklin:
The upstream Drupal module may be able to send it as an int (and I believe there is already a patch that does that), but this library should also validate the data it receives, and ensure the JSON transactions (which Drupal knows nothing about) are formatted according to spec, even when data is sent (implicitly) cast to another data type.

PHP is very loose with types, recasting scalar values as needed, and so most PHP developers don't give specific scalar types much thought, anymore than they give memory management much thought.  It's only when something else demands a specific scalar type that it becomes an issue.  

I think this is a case where Postel's law should be applied.

PR #941, opened 2018-07-13
Issue #940: Explictly cast 'expires_in' to an int to prevent returning a JSON String.

jlfranklin:


bshaffer:
Can you explain why this is necessary? Thanks.

danopz:
ref #940 

Looks like this is just a config value provided from the dev? If thats the case this shouldn't be neccessary because you should just set an int value in the config.

Issue #924, opened 2018-03-27
Incorrect response code for user credentials grant type


SventB:
The status code 401 introduced in #290 was wrong, it should be 400, see https://tools.ietf.org/html/rfc6749#section-5.2, but see especially https://stackoverflow.com/questions/22586825/oauth-2-0-why-does-the-authorization-server-return-400-instead-of-401-when-the and https://www.ietf.org/mail-archive/web/oauth/current/msg02127.html

PR #962, opened 2019-06-04
Modified response code 400 to 401

whvandervelde:
When passing invalid client credentials (either client_id or client_secret) when requesting an access token, instead of returning a 400 response code, it should be returning a 401 response code for unauthorized. Limited it to the case where all conditions (public/non public etc) are passed and its only about the credentials. TokenControllerTest modified accordingly.

### DIFF
--- a/src/OAuth2/ClientAssertionType/HttpBasic.php
+++ b/src/OAuth2/ClientAssertionType/HttpBasic.php
@@ -71,7 +71,7 @@ class HttpBasic implements ClientAssertionTypeInterface
                 return false;
             }
         } elseif ($this->storage->checkClientCredentials($clientData['client_id'], $clientData['client_secret']) === false) {
-            $response->setError(400, 'invalid_client', 'The client credentials are invalid');
+            $response->setError(401, 'invalid_client', 'The client credentials are invalid');
 
             return false;
         }
--- a/test/OAuth2/Controller/TokenControllerTest.php
+++ b/test/OAuth2/Controller/TokenControllerTest.php
@@ -96,7 +96,7 @@ class TokenControllerTest extends TestCase
         ));
         $server->handleTokenRequest($request, $response = new Response());
 
-        $this->assertEquals($response->getStatusCode(), 400);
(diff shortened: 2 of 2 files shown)


CameronGo:
Just uncovered this in our environment today. Any idea when this PR may get merged or why not?

CameronGo:
Just saw the note from another issue indicating that the 400 response is the desired and correct response code for this scenario. https://github.com/bshaffer/oauth2-server-php/pull/846

I'm not an expert in this, but I believe the RFC allows for a 401 response here based on this note: https://tools.ietf.org/html/rfc6749#section-6

Then here (which I realize is not the RFC) specifically states that an invalid client ID or client secret will return a 401:
https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/

whvandervelde:
Hey, wow this has been a while ago ;-)
Yes, this is specifically (only) done for Client Credentials where I think 401 would be more appropriate, as per your reference of the RFC. On a general note I think it just makes sense on what a 401 should be used for vs a 400 and to able to provide better error handling for anybody using this.

There was some discussion where 400 was the preferred choice with User Credentials because 401 requires a challenge in the response, see #924. What you refer too with #846 is for a refresh_token, that probably also still would be a 400 preferred?

whvandervelde:
The reason I prefer a 401 here is because we want to use a specific flow for a client in case of supplying invalid credentials and we don't want to tie that to a specific implementation's error message content.

Spomky:
The error 400 is the correct error code as per the RFC. There is no reason to change it.

CameronGo:
@Spomky  - i'm not saying the 400 response is wrong per the RFC, but from the links provided it seems to be the RFC _allows_ for a 401 as well. The additional link I provided shows an example from oauth.com where they in fact _are_ returning a 401 in the specific scenario described here.

So while I don't think it _has_ to be changed, it seems to me that it could be changed and still be correct and would be helpful for anyone attempting to code specifically for invalid credentials, which seems like a totally reasonable thing to do.

PR #965, opened 2019-06-27
OpenIdConnect: Add scopes parameter to enable claims in id-token

Ben-Ho:


bshaffer:
Hi @Ben-Ho ! Thank your for your PR
Can you explain a little but more the implications of the PR?

bshaffer:
At the very least, `UserClaimsStorage` needs to be optional in the constructor of `OAuth2\OpenID\Controller\AuthorizationController`, and we need to check it exists before calling `getUserClaims`.  We also need some tests to verify the functionality.

Thank you!

Ben-Ho:
currently I'm not sure if my PR is really a good idea as I found https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.5.4. There is this sentence: 

> However, when no Access Token is issued (which is the case for the response_type value id_token), the resulting Claims are returned in the ID Token.

And the only benefit is it reduces the requirement of one http-request to user-info endpoint (but only for code-flow)

This was also discussed for a different OAuth-Server project https://github.com/IdentityServer/IdentityServer4/issues/528.

PR #979, opened 2020-03-17
URL-decode client credentials in HTTP Basic auth, as described in the RFC

pjcdawkins:
https://tools.ietf.org/html/rfc6749#section-2.3.1
> Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server.  The client identifier is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm per
   Appendix B, and the encoded value is used as the username; the client
   password is encoded using the same algorithm and used as the
   password.

This affects client IDs and secrets that would differ from their URL-decoded form, i.e. those containing a sequence starting with `%`.


bshaffer:
This is a great change but as it would break backwards compatibility we'll need to save this for the next major version

PR #1018, opened 2022-03-23
chore: migrate to gh actions

bshaffer:


snapshotpl:
You can try https://github.com/Yoast/PHPUnit-Polyfills to fix issues on older PHP versions

snapshotpl:
Or we should consider drop old PHP version support

trylika:
we are talking about php 5.6/7.0/7.1 versions, which are currently ancient, i would definitely suggest making life easier by dropping them

kristijonasb:
+1 for dropping support for older versions.

trylika:
@bshaffer any news?

bshaffer:
Okay I dropped support for `7.0` as well, and we should be able to merge this soon. 

One question though... why is this a priority to the people on this thread? This does not add support or compatibility, it only affects the test suite. So I'm confused how something could be reliant on any of the changes being made here.

Thanks!

PR #1021, opened 2022-05-11
Fix PHP 8.1 deprecation error

snapshotpl:
Found when test on PHP 8.1 here https://github.com/laminas-api-tools/api-tools-oauth2/runs/6385928373?check_suite_focus=true#step:3:677


src/OAuth2/TokenType/Bearer.php
```diff
@@ -109,7 +109,7 @@ public function getAccessTokenParameter(RequestInterface $request, ResponseInter
             }
 
             $contentType = $request->server('CONTENT_TYPE');
-            if (false !== $pos = strpos($contentType, ';')) {
+            if (is_string($contentType) && false !== $pos = strpos($contentType, ';')) {
```
Ocramius, line 112: Is it possible to test this?

Ocramius (changes requested):

src/OAuth2/TokenType/Bearer.php
snapshotpl: Fixed and better fix applied. Thanks!

Ocramius: Awesome, thanks!

test/OAuth2/TokenType/BearerTest.php
```diff
@@ -35,6 +35,20 @@ public function testInvalidContentType()
...
+        $bearer = new Bearer();
+        $request = TestRequest::createPost(array(
+            'access_token' => 'ThisIsMyAccessToken'
+        ));
+
+        $param = $bearer->getAccessTokenParameter($request, $response = new Response());
+        $this->assertNull($param);
+        $this->assertEquals($response->getStatusCode(), 400);
+        $this->assertEquals($response->getParameter('error'), 'invalid_request');
+        $this->assertEquals($response->getParameter('error_description'), 'The content type for POST requests must be "application/x-www-form-urlencoded"');
```
Ocramius, line 49: Non-blocker, but left is expected, right is actual, IIRC 🤔

Ocramius (approved):

test/OAuth2/TokenType/BearerTest.php
snapshotpl: In my own defense I copied it from another test case.

bshaffer (approved):

PR #1022, opened 2022-05-11
PHPStan level 0

snapshotpl:


bshaffer (approved):

.github/workflows/tests.yml
```diff
@@ -53,3 +53,19 @@ jobs:
           command: composer install
       - name: Run PHPUnit
         run: vendor/bin/phpunit -v
+  phpstan:
+    name: "PHPStan"
```
bshaffer, line 57: The actions are erroring out with the following message:
```
The workflow is not valid. .github/workflows/tests.yml (Line: 57, Col: 5): Required property is missing: runs-on
```

```suggestion
    name: "PHPStan"
    runs-on: ubuntu-latest
```

bshaffer:
The tests are failing because `phpstan` is not compatible with 7.1. We don't need it to be, as we only run it on one PHP version. So we should remove it from the dev dependencies and add `phpstan` at test time so that we don't get the PHP 7.1 error. Something like:

```
run: |
  composer require phpstan/phpstan
  vendor/bin/phpstan
```

bshaffer (approved):

PR #1036, opened 2022-11-14
Add support for xml request , to add xml request into Request object

Salman98:
While working with XML Request when I pass the request as XML from postman. 
The lib Request class unable to read the Request parameters. After adding these changes I was able to continue.


bshaffer:
Hello! Thank you for submitting this! Could you please add a test?

PR #1044, opened 2023-02-21
upgrade firebase/jwt to v6

### LINKED ISSUE #1043: Firebase/JWT <6 is considered security risk
Update the recommended version and fix the code. The new versions of Firebase/JWT has namespaced classes.

jasverix:
Fix #1043

### DIFF
--- a/src/OAuth2/Storage/JwtAccessToken.php
+++ b/src/OAuth2/Storage/JwtAccessToken.php
@@ -21,7 +21,7 @@ class JwtAccessToken implements JwtAccessTokenInterface
      *                                                                is not necessary when using this grant type.
      * @param OAuth2\Encryption\EncryptionInterface $encryptionUtil   OPTIONAL class to use for "encode" and "decode" functions.
      */
-    public function __construct(PublicKeyInterface $publicKeyStorage, AccessTokenInterface $tokenStorage = null, EncryptionInterface $encryptionUtil = null)
+    public function __construct(PublicKeyInterface $publicKeyStorage, AccessTokenInterface $tokenStorage, EncryptionInterface $encryptionUtil = null)
     {
         $this->publicKeyStorage = $publicKeyStorage;
         $this->tokenStorage = $tokenStorage;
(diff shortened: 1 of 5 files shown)


jasverix:
This branch now works, but a *breaking change* is that the TokenStorage is no longer mandatory in JwtAccessToken storage.

src/OAuth2/Storage/JwtAccessToken.php
```diff
@@ -34,11 +34,11 @@ public function __construct(PublicKeyInterface $publicKeyStorage, AccessTokenInt
     public function getAccessToken($oauth_token)
     {
         // just decode the token, don't verify
-        if (!$tokenData = $this->encryptionUtil->decode($oauth_token, null, false)) {
+        if (!$tokenData = $this->tokenStorage->getAccessToken($oauth_token)) {
```
bshaffer, line 37: Why does this need to be changed in order to support the latest version of `firebase/php-jwt`?

src/OAuth2/Storage/JwtAccessToken.php
```diff
@@ -34,11 +34,11 @@ public function __construct(PublicKeyInterface $publicKeyStorage, AccessTokenInt
...
             return false;
         }
 
-        $client_id  = isset($tokenData['aud']) ? $tokenData['aud'] : null;
+        $client_id  = isset($tokenData['client_id']) ? $tokenData['client_id'] : null;
```
bshaffer, line 41: Similarly, I think this should not need to change

```suggestion
        $client_id  = isset($tokenData['aud']) ? $tokenData['aud'] : null;
```

src/OAuth2/Storage/JwtAccessToken.php
jasverix: It is some time ago, I do not really remember. But I think there was a problem that the newest firebase needed the key always, and in this if-test, `null` was passed.

The idea here is to decode the token data passed, and in my project at least, the tokenStorage had all information I needed. But maybe you should do it in 2 steps then, first get data from tokenStorage (I think it contains the key) and then decode with the encryptionUtil.

But if I don't remember wrong, I think it is not possible to pass `null` to the decode method if firebase is used.

src/OAuth2/Storage/JwtAccessToken.php
jasverix: This is because of the other change. When fetched from the tokenStorage, the client_id is called client_id, not aud. If you revert one of them, you need to revert both.

I actually think that when I did this change, I first started by getting the data from the tokenStorage to be able to find the key to pass into the decode() method. But then I saw that it was only used to fetch the client_id, and that we also have it in the tokenStorage data, so I just decided to use that instead.

PR #951, opened 2018-11-20
Add support for OpenID Connect PKCE

darthsteven:
Very much a WIP and not finished, just getting something out there and started.

Stay tuned!

PR #1045, opened 2023-02-26
Add support of PKCE

hkalina:
This is built on https://github.com/bshaffer/oauth2-server-php/pull/951, but missing parts was added:
* Updated Storage classes to support new code_challenge/code_challenge_method fields
* Added support of PKCI even into non-openid Oauth2
* Tested using modified oauth2-demo-php: https://github.com/hkalina/oauth2-demo-php/tree/pkce

Any feedback is welcome!

### DIFF
--- a/src/OAuth2/GrantType/AuthorizationCode.php
+++ b/src/OAuth2/GrantType/AuthorizationCode.php
@@ -84,6 +84,45 @@ class AuthorizationCode implements GrantTypeInterface
             return false;
         }
 
+        // @TODO: Should we enforce presence of a non-falsy code challenge?
+        if (isset($authCode['code_challenge']) && $authCode['code_challenge']) {
+          if (!($code_verifier = $request->request('code_verifier'))) {
+            $response->setError(400, 'code_verifier_missing', "The PKCE code verifier parameter is required.");
+
+            return false;
+          }
+          // Validate code_verifier according to RFC-7636
+          // @see: https://tools.ietf.org/html/rfc7636#section-4.1
+          elseif (preg_match('/^[A-Za-z0-9-._~]{43,128}$/', $code_verifier) !== 1) {
+            $response->setError(400, 'code_verifier_invalid', "The PKCE code verifier parameter is invalid.");
+
+            return false;
+          }
+          else {
+            $code_verifier = $request->request('code_verifier');
+            switch ($authCode['code_challenge_method']) {
+              case 'S256':
+                $code_verifier_hashed = strtr(rtrim(base64_encode(hash('sha256', $code_verifier, true)), '='), '+/', '-_');
+                break;
+
+              case 'plain':
+                $code_verifier_hashed = $code_verifier;
+                break;
+
+              default:
+                $response->setError(400, 'code_challenge_method_invalid', "Unknown PKCE code challenge method.");
+
+                return FALSE;
+            }
+            // @TODO: use hash_equals in recent versions of PHP.
+            if ($code_verifier_hashed !== $authCode['code_challenge']) {
+              $response->setError(400, 'code_verifier_mismatch', "The PKCE code verifier parameter does not match the code challenge.");
+
+              return FALSE;
+            }
+          }
+        }
+
         if (!isset($authCode['code'])) {
             $authCode['code'] = $code; // used to expire the code after the access token is granted
         }
--- a/src/OAuth2/GrantType/AuthorizationCode.php
+++ b/src/OAuth2/GrantType/AuthorizationCode.php
@@ -84,7 +84,6 @@ class AuthorizationCode implements GrantTypeInterface
             return false;
         }
 
-        // @TODO: Should we enforce presence of a non-falsy code challenge?
         if (isset($authCode['code_challenge']) && $authCode['code_challenge']) {
           if (!($code_verifier = $request->request('code_verifier'))) {
             $response->setError(400, 'code_verifier_missing', "The PKCE code verifier parameter is required.");
--- a/src/OAuth2/GrantType/AuthorizationCode.php
+++ b/src/OAuth2/GrantType/AuthorizationCode.php
@@ -85,41 +85,38 @@ class AuthorizationCode implements GrantTypeInterface
         }
 
         if (isset($authCode['code_challenge']) && $authCode['code_challenge']) {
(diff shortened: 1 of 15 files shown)


src/OAuth2/GrantType/AuthorizationCode.php
```diff
@@ -84,6 +84,44 @@ public function validateRequest(RequestInterface $request, ResponseInterface $re
...
 
+        if (isset($authCode['code_challenge']) && $authCode['code_challenge']) {
+          if (!($code_verifier = $request->request('code_verifier'))) {
+            $response->setError(400, 'code_verifier_missing', "The PKCE code verifier parameter is required.");
+
+            return false;
+          }
+          // Validate code_verifier according to RFC-7636
+          // @see: https://tools.ietf.org/html/rfc7636#section-4.1
+          elseif (preg_match('/^[A-Za-z0-9-._~]{43,128}$/', $code_verifier) !== 1) {
```
bshaffer, line 95: Implicit else!
```suggestion
          if (preg_match('/^[A-Za-z0-9-._~]{43,128}$/', $code_verifier) !== 1) {
```

src/OAuth2/GrantType/AuthorizationCode.php
```diff
@@ -84,6 +84,44 @@ public function validateRequest(RequestInterface $request, ResponseInterface $re
...
+            $response->setError(400, 'code_verifier_invalid', "The PKCE code verifier parameter is invalid.");
+
+            return false;
+          }
+          else {
```
bshaffer, line 100: Implicit else here too, you can remove the `else`

src/OAuth2/GrantType/AuthorizationCode.php
```diff
@@ -84,6 +84,44 @@ public function validateRequest(RequestInterface $request, ResponseInterface $re
...
+                return FALSE;
+            }
+            // @TODO: use hash_equals in recent versions of PHP.
+            if ($code_verifier_hashed !== $authCode['code_challenge']) {
+              $response->setError(400, 'code_verifier_mismatch', "The PKCE code verifier parameter does not match the code challenge.");
+
+              return FALSE;
+            }
+          }
+        }
```
bshaffer, line 123: Whitespace in this whole section is off, please use four spaces!

bshaffer (commented):
This is a great PR! Thank you for adding PKCE support! Just some minor formatting changes and we can merge this.

I'll also add tests in another PR unless you'd like to add those as well

hkalina:
@bshaffer Updated, thank you! :)

jasverix:
Should probably mark this upgrade somehow, because it breaks the system if the developers does not apply a migration. I don't have a good suggestion on how you could have marked it, but when I first looked at the upgrade log, I did not detect that it was a breaking upgrade.

I will of course just do the migration and continue to use this package, but I just wanted to give this feedback. It is a very well working and good package.

bshaffer:
This also SHOULDNT be a breaking change, it should be opt-in. Can you post the error you received as well?

bshaffer:
Great. I figured it was something like that. That shouldn’t be too hard to fix then.

Wirone:
This is a breaking change because of change in `AuthorizationCodeInterface::setAuthorizationCode()` signature. Any class implementing this [causes fatal error](https://3v4l.org/JOtqn) now. In our case Renovate Bot tried to upgrade this package, but fortunately our tests failed, so it wasn't merged.

image

@bshaffer what's the plan for it? Consider using [PHP SemVer Checker](https://github.com/tomzx/php-semver-checker) 🙂.

Wirone:
@bshaffer what's the situation? We have internal MR with changes aligning our code with this change, but we don't know if we should merge it because if revert is done, we also would need to revert on our side 🙈. Please let us know.

bshaffer:
I am sorry, I am not really maintaining this library more than merging PRs. If you have a fix you'd like to submit I'm happy to review it.

Wirone:
@bshaffer I was just referring to [what you said](https://github.com/bshaffer/oauth2-server-php/pull/1045#issuecomment-1591264454):

> I will revert it

But it never happened 😉. Since it's a change in the interface, there's no way to refactor it in non-breaking way (moving those 2 added arguments to separate method would be breaking as well). I believe it's a change worth keeping, so what should be done IMHO:

- revert this change and tag 1.14.1: people who did not upgrade yet would have clean upgrade path, without breaking change
- bring back this change and tag 2.0.0: people who already upgraded to 1.14.0 would have another BC break, but they should only bump to 2.0 and continue with their current code (that's why there shouldn't be other BC-breaks in 2.0, to be compatible with what people already did).

We just need clear information what's the plan about this, if it's going to be kept as-is, we'll just merge changes on our side and continue with this BC-break. If you're going to revert it, we won't change implementation on our side 🙂.

bshaffer:
IMHO adding two empty methods to whatever custom classes extend the interface in a dependent library is not too much to ask. I understand that this breaks semver, but at this point the damage is done, and someone can simply pin to the previous version if they don't want the breaking change.

rdamas:
The documentation here: https://bshaffer.github.io/oauth2-server-php-docs/cookbook/ should probably be updated to reflect the database change, the columns `code_challenge` and `code_challenge_method` are not mentioned yet.

aarangara:
Is this (PKCE) only supposed to work with OpenID - because currently it does not seem to work for me when `use_openid_connect = false`.

For e.g. The Oauth2\Controller\AuthorizeController does not have any code that set-ups the code_challenge and code_challenge_variable. The only change this MR had to that file is `'enforce_pkce' => false`

But if you check Oauth2\OpenID\Controller\AuthorizeController, it has got some code in `validateRequest` does some checks for PKCE.

campbsb:
> The documentation here: https://bshaffer.github.io/oauth2-server-php-docs/cookbook/ should probably be updated to reflect the database change, the columns `code_challenge` and `code_challenge_method` are not mentioned yet.

+1

I've had to add best-guess definitions for these undocumented code_challenge and code_challenge_method fields.

bshaffer:
Thanks for the nudge, I will update it

ArnoHolo:
I suggest you indicate it as a Breaking Change in the Release of the 1.14.0 version => https://github.com/bshaffer/oauth2-server-php/releases/tag/v1.14.0

bshaffer:
@ArnoHolo done

Issue #1056, opened 2023-11-22
Not possible to inject custom implementation of EncryptionInterface without overriding whole createDefaultIdTokenResponseType() or whole response type


A-Shevchenko:
Let's imagine we want to have custom implementation of EncryptionInterface, e.g. because we want to add custom header info. It's the instruction directly in the code BTW:
```
    /**
     * Override to create a custom header
     */
    protected function generateJwtHeader($payload, $algorithm)
    {
        return array(
            'typ' => 'JWT',
            'alg' => $algorithm,
        );
    }
```
So, we create a custom implementation, but there is no way to pass it to instantiation of `IdToken` class, because it doesn't accept anything as last parameter here: `return new IdToken($this->storages['user_claims'], $this->storages['public_key'], $config);` (in `createDefaultIdTokenResponseType()`).
So to solve that we need to create a custom implementation of createDefaultIdTokenResponseType() or whole response type

PR #1057, opened 2023-11-24
added possibility to inject custom encryption object for IdToken response type

A-Shevchenko:
This is to address issue #1056 - simple dependency injection via setter + getter to fallback to default object if none was explicitly passed


A-Shevchenko:
Also added the same thing for case with JWT access tokens

PR #1067, opened 2024-06-07
Add support for authorization code grant with PKCE for distributed apps

ls-sean-fraser:
Add support for "Authorization Code Grant with PKCE" flow to allow distributed apps to securely authenticate with an OAuth server without needing to embed secrets within the distributed application, as these can be easily exposed by decompiling the application binaries.

- Move handlers for code challenge from OIDC AuthorizationController to the base AuthorizationController, so that non-OIDC flows can leverage it as well
- Add mechanism to enforce PKCE code challenges for public clients
- Add mechanism to configure supported code challenge methods
- Change code_verifier comparison to use hash_equals() to avoid timing attacks
- Added tests for all PKCE code challenge flows

Reference:
- [RFC-8252](https://datatracker.ietf.org/doc/html/rfc8252)

### DIFF
--- a/src/OAuth2/Controller/AuthorizeController.php
+++ b/src/OAuth2/Controller/AuthorizeController.php
@@ -2,6 +2,7 @@
 
 namespace OAuth2\Controller;
 
+use OAuth2\Storage\ClientCredentialsInterface;
 use OAuth2\Storage\ClientInterface;
 use OAuth2\ScopeInterface;
 use OAuth2\RequestInterface;
@@ -62,6 +63,16 @@ class AuthorizeController implements AuthorizeControllerInterface
     protected $scopeUtil;
 
     /**
+     * @var string|null
+     */
+    protected $code_challenge;
+
+    /**
+     * @var string|null
+     */
+    protected $code_challenge_method;
+
+    /**
      * Constructor
      *
      * @param ClientInterface $clientStorage REQUIRED Instance of OAuth2\Storage\ClientInterface to retrieve client information
@@ -88,6 +99,8 @@ class AuthorizeController implements AuthorizeControllerInterface
             'require_exact_redirect_uri' => true,
             'redirect_status_code' => 302,
             'enforce_pkce' => false,
+            'enforce_pkce_for_public_clients' => false,
+            'supported_code_challenge_methods' => ['plain', 'S256'],
         ), $config);
 
         if (is_null($scopeUtil)) {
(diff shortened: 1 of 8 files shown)


src/OAuth2/Controller/AuthorizeController.php
```diff
@@ -88,6 +99,8 @@ public function __construct(ClientInterface $clientStorage, array $responseTypes
             'require_exact_redirect_uri' => true,
             'redirect_status_code' => 302,
             'enforce_pkce' => false,
+            'enforce_pkce_for_public_clients' => false,
```
ls-sean-fraser, line 102: ℹ️ this should probably default to true for new installs, but that could be a breaking change for existing installs. Leaving this at false allows for releasing as a minor version. Would recommend defaulting to true for the next major version.

src/OAuth2/Controller/AuthorizeController.php
```diff
@@ -334,13 +349,50 @@ public function validateAuthorizeRequest(RequestInterface $request, ResponseInte
...
+            if (!in_array($code_challenge_method, $supported_challenge_methods, true)) {
+                $response->setError(400, 'invalid_code_challenge_method', 'The PKCE code challenge method supplied is invalid. Supported methods: ' . implode(', ', $supported_challenge_methods)
+                );
+
+                return false;
+            }
+        } else {
+            $isPublic = false;
+            if ($this->clientStorage instanceof ClientCredentialsInterface) {
+                $isPublic = $this->clientStorage->isPublicClient($client_id);
```
ls-sean-fraser, line 377: ℹ️ not sure why there are two interfaces, but the property is only typed to the higher level one which doesn't have the method we need.

src/OAuth2/OpenID/Controller/AuthorizeController.php
```diff
@@ -16,16 +16,6 @@ class AuthorizeController extends BaseAuthorizeController implements AuthorizeCo
      */
     private $nonce;
 
-    /**
-     * @var mixed
-     */
-    protected $code_challenge;
```
ls-sean-fraser, line 22: ℹ️ these properties and their behavior moved to the parent class

src/OAuth2/Server.php
```diff
@@ -173,6 +173,8 @@ public function __construct($storage = array(), array $config = array(), array $
             'require_exact_redirect_uri' => true,
             'allow_implicit'           => false,
             'enforce_pkce'             => false,
+            'enforce_pkce_for_public_clients' => false,
```
ls-sean-fraser, line 176: ℹ️ same comment about default value

ls-sean-fraser:
@bshaffer any chance you could give this PR a glance?

bshaffer:
@ls-sean-fraser will do!

PR #1076, opened 2025-02-07
Fix PHP 8.4 syntax

afridlund85:
Just ran the php-cs-fixer with the `nullable_type_declaration_for_default_null_value` rule to make the code 8.4 compatible.


afridlund85:
Saw now that this already had a PR pending.

PR #1090, opened 2026-04-24
Use of "self" in callables is deprecated in PHP 8.1

MarkMaldaba:
There were a number of places in the DynamoDB class that use 'self::isNotEmpty' as a callback function.  However, relative callables (e.g. using `self`, `parent`, etc.) have been deprecated in PHP 8.1 and will result in a deprecation notice being triggered.  I have updated these to use `__CLASS__` instead.

Reference: https://www.php.net/manual/en/migration82.deprecated.php#migration82.deprecated.core.relative-callables


MarkMaldaba:
Apologies for the newline at the end of the file.  It was inserted by the Github editor and although I've tried to remove it, it doesn't seem to be possible without checking out the repo.  I hope someone can fix this prior to merging.

